PR #12, opened 2026-02-23
Add functions to resolve DOIs across multiple providers

caufieldjh:
This pull request introduces DOI ingestion support for multiple data repositories by adding specialized resolver modules. Each resolver implements logic to retrieve and normalize abstract or description metadata for a given DOI from its respective source. Additionally, shared parsing helpers are introduced to streamline and standardize text extraction and error handling across resolvers.

### DIFF
--- a/tests/test_doi_ingestion.py
+++ b/tests/test_doi_ingestion.py
@@ -20,6 +20,7 @@ from urllib.parse import parse_qs, urlparse
 
 import pytest
 import responses
+from requests import PreparedRequest
 
 from nmdc_metadata_suggestor.constants import (
     CROSSREF_API_URL as CROSSREF_API,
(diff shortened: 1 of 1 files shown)


turbomam (commented):
Nice. I have two requests and one suggestion, but I wouldn't block the PR on any of them.
                                                                                                                                                                                             
1. Standardize the resolver return types. `try_figshare` returns `str`, `try_ess_dive` returns `tuple[str, str]`, `try_kbase` returns `tuple[str, str, str]`. A common NamedTuple with text, raw_text, kind fields would simplify the `_fetch_*` wrappers in `main.py` and make adding future resolvers less error-prone.
2. MassIVE resolver silently calls DataCite.  `_extract_massive_context_from_datacite_titles()` hits DataCite as a fallback but still reports `source="massive"`. Either report the actual source or document why this is intentional.

Suggestion: more  integration tests for the newer resolvers (CyVerse, EDI especially) would help catch API response drift that fixture mocking won't.

hesspnnl (approved):

PR #19, opened 2026-03-03
Build pattern with Osti for pub link in waterfall

hesspnnl:
Add checking for publication links to the main context of the waterfall

### DIFF
--- a/tests/test_abstract_retriever.py
+++ b/tests/test_abstract_retriever.py
@@ -256,7 +256,10 @@ class TestGetAbstractWaterfall:
 
     @responses.activate
     def test_osti_link_only_continues_to_crossref_and_accumulates_metadata(self) -> None:
-        """OSTI link-only result should not stop waterfall; Crossref text should still be returned."""
+        """
+        OSTI link-only result should not stop waterfall
+        Crossref text should still be returned.
+        """
         requested_doi = "10.15485/7654321"
         publication_doi = "10.1093/nar/gkac972"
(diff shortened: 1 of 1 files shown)


src/nmdc_metadata_suggestor/doi_ingestion/osti.py
```diff
@@ -42,17 +37,40 @@ def try_osti(doi: str, errors: list[str] | None = None) -> ResolverContext | Non
...
 
     record = data[0] if isinstance(data, list) else data
-    abstract = record.get("description")
-    if not isinstance(abstract, str):
-        append_error(errors, "No abstract/description found in OSTI record")
-        return None
 
-    cleaned = clean_text(abstract)
-    if not cleaned:
+    raw_description = record.get("description")
```
caufieldjh: It's possible that the record will be retrievable and valid but won't contain much useful text (e.g., https://www.osti.gov/biblio/2423571). In an edge case it may not even be present, and that gets caught in lines 45-46, but it will still be returned as a `ResolverContext` if it gets a PDF link. That might be OK if other sources also don't provide additional text.

caufieldjh (approved):
I think this is ready to go, but one of the next steps will need to be a broader evaluation than the current integration tests are doing. I'm thinking about at least a few hundred DOIs (maybe from those already used in NMDC entries) to catch edge cases and publication sets unexpectedly missing from one source or another.

PR #35, opened 2026-03-11
Add model logging, remove claude tests, update google auth

hesspnnl:
In this PR i added the model name and access provider to the output dict. I also needed to update the google auth flow to support credential file via Oauth (vertex specific). I also created a max tokens dict for the different models/providers and removed old claude test.


turbomam (commented):
Tested with my GCP credentials (`service-account.json`) — auth initializes and `generate()` returns a response from Gemini 2.0 Flash. The ADC fallback refactor looks solid.

One request: the `__main__` block in `recommendation_pipeline.py` hardcodes `access_provider="pnnl"`, so I had to write a throwaway script to test with GCP. Could we add `--access-provider` as a CLI arg (or at least an env var check) so contributors can smoke-test with either backend without editing code?

Minor notes:
- `DEFAULT_MAX_TOKENS_BY_PROVIDER["pnnl"] = 128000` — is that intentional? 128k looks like context window size, not max output tokens. GCP's 8192 seems right for output.
- `generate()` renamed `temperature` to `gemini_temperature` and dropped it entirely from the PNNL path — worth noting as a breaking change for any existing callers.

PR #38, opened 2026-03-16
Expand publication metadata retrieval to additional providers

caufieldjh:
This PR adds additional functionality to the DOI-based metadata retrieval to obtain publication details from additional providers.
(See https://github.com/microbiomedata/issues/issues/1632)

For some providers (e.g., Zenodo and Figshare), publication metadata is available in structured data.
For EMSL and MassIVE, I added additional resolvers to search additional locations for publication metadata.
For EDI, JGI, CyVerse, and KBase, I was not able to identify a consistent metadata location for linked publications. There are functions included in this PR to parse loosely structured text fields where this metadata *may* be, but these functions are currently disabled as they aren't testable without some concrete working examples.

It also updates the README - please see the updated doc for details regarding what we can retrieve through each provider.


hesspnnl (changes requested):
I think in general this looks good - butI think for the providers that do not directly have "pdf link" in their metadata - we should not add more code and instead should add a call to openalex or pubmed to check for a publication. Ie edi, emsl, jgi, kbase, cyverse. We can put the code you built out for string matching etc on another branch and keep it in draft until we decide what to do with it. @turbomam feel free to chime in with your thoughts as well

caufieldjh:
Closing as stale - will try to salvage some material in the future if we truly require more metadata from individual providers, but let's confirm that's necessary first.

PR #16, opened 2026-03-02
Add Elsevier ScienceDirect API as waterfall step 2.5

turbomam:
## Summary

Closes https://github.com/microbiomedata/issues/issues/1611

Elsevier withholds abstracts from Crossref, so the existing publication waterfall (OpenAlex → Crossref → PubMed → content negotiation → OSTI) failed for most `10.1016` DOIs (421 in NMDC inventory). This adds the ScienceDirect Abstract Retrieval API as step 2.5 between Crossref and PubMed.

- **New resolver:** `doi_ingestion/elsevier.py` with `try_elsevier()` and `try_elsevier_abstract()`
- **Prefix-gated:** Only calls the API for `10.1016/` DOIs; skips instantly for other prefixes (no wasted API calls)
- **Graceful degradation:** When `ELSEVIER_API_KEY` is absent, the step is silently skipped
- **API key read at call time** via `os.environ.get()`, not at import time (avoids issues with `.env` loading order)

### Waterfall order (updated)
1. OpenAlex → 2. Crossref → **2.5. Elsevier** → 3. PubMed → 4. Content negotiation → 5. OSTI

### Files changed
| File | Change |
|------|--------|
| `constants.py` | `ELSEVIER_API_URL`, `"elsevier"` in `ALL_SOURCES` |
| `doi_ingestion/elsevier.py` | **New** — resolver module |
| `abstract_retriever.py` | Wire `_fetch_elsevier` into `_SOURCE_FETCHERS` |
| `doi_response_payloads.json` | Elsevier fixture payloads |
| `test_abstract_retriever.py` | 6 unit tests + 2 integration tests |
| `doi-classification-design.md` | Waterfall table, env vars, API key guidance |
| `.env.example` | `ELSEVIER_API_KEY` placeholder |

## Test plan

- [x] `uv run pytest` — 167 passed, 0 failures (all existing tests updated for new waterfall order)
- [x] `uv run pytest -k elsevier -v` — 6 unit tests pass
- [x] `uv run pytest -m integration -k elsevier -v` — 3 integration tests pass (with API key)
- [x] `make get-abstract DOI=10.1016/j.apsoil.2025.106110 SOURCES=elsevier` — returns abstract via Elsevier source
- [x] `make get-abstract DOI=10.1038/s41564-020-00861-0` — Elsevier skipped (non-10.1016), works via OpenAlex

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #65, opened 2026-04-03
Add Elsevier and Springer Nature as waterfall sources

turbomam:
## Summary
- Add `elsevier.py` and `springer_nature.py` DOI ingestion modules (ported from #16, updated for package rename)
- Wire both into the source fetcher waterfall in `main.py`
- Add API URL constants and `.env.example` entries for API keys
- Update `ALL_SOURCES` and its test

Supersedes #16 (which targeted the old `nmdc_metadata_suggestor` package path).

## Test plan
- [x] All 191 tests pass (35 deselected integration tests)
- [x] `ruff check` and `ruff format --check` pass
- [ ] Integration test with real Elsevier/Springer API keys

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -27,6 +27,15 @@ CONTACT_EMAIL=support@microbiomedata.org
 # Request at https://docs.openalex.org/how-to-use-the-api/api-key
 # OPENALEX_API_KEY=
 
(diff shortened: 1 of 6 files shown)


turbomam:
Closing — we decided to stick to open-access sources only. See #68 for tracking.

PR #71, opened 2026-04-06
Split out ENV triad filling

### LINKED ISSUE #53: Split out ENV triad filling and slot recommendation
The ETL scripts have a separate need for ENV triad value filling. These two processes are coupled right now - but need to be split out into two pipelines that can be called separately. 

Things to keep in mind:

- code changes need to be minimal - the recommendation pipeline input and output should stay the same to save code changes on the server
- We will need a new system prompt for just env triad filling
- we will need the ability to just pass the env triad schema enums to the env triad filling function
- inputs to the env triad value filling function should be agnostic so multiple parts of NMDC can call it
- The LLMClient may need changes as right now it is a set system prompt - taking as input the task the LLMClient is completing may be a good solution. that way sys prompts can still live on the ai tool code.

hesspnnl:
Split out the env triad into its own module.


src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
```diff
@@ -45,3 +40,35 @@
 }
 ```
 """
+
+
+env_triad_prompt = """
```
caufieldjh, line 45: Should this just be handled as a structured output, i.e., explicitly require JSON output validated against the schema?
https://developers.openai.com/api/docs/guides/structured-outputs
I admittedly haven't used structured outputs much because I don't always know what the input schema is going to look like in advance, but in this case we *just* want JSON and nothing else.

tests/test_env_triad_recommendation.py
```diff
@@ -0,0 +1,95 @@
...
+    field_names = {f.field_name for f in recommended_metadata.metadata_fields}
+    assert env_triad_names & field_names, f"Expected env triad field in {field_names}"
+
+
+def test_run_env_triad_pipeline_with_biosample(requires_credentials: None) -> None:
+    llm_client = LLMClient(access_provider="gcp")
+    recommended_metadata = env_triad_recommendation.get_env_triad_recommendation(
+        context=[
+            {
+                "biosample_set": [
```
caufieldjh: Could this biosample set be separated into a test fixture? It looks like it may be relevant to other tests.

caufieldjh (approved):
Looks ready - just a couple comments

src/nmdc_metadata_suggestor_ai_tool/schema_context.py
```diff
@@ -205,7 +202,7 @@ def filter_slots(self, schema: InterfaceSchemaClass) -> InterfaceSchemaClass:
         for slot in schema.slots:
             if slot.name in EXCLUDED_SLOTS:
                 continue
-            if (slot.required or slot.deprecated) and slot.name not in INCLUDED_REQUIRED_SLOTS:
+            if (slot.required or slot.deprecated) and slot.name:
```
turbomam: what' the value of checking the truthiness of `slot.name` here? Isn't it always `true`?

tests/test_llm_client.py
```diff
@@ -20,10 +21,11 @@ def test_generate_uses_pnnl_default_max_tokens_when_not_provided(monkeypatch: An
     client = LLMClient.__new__(LLMClient)
     client.access_provider = "pnnl"
     client.model = "gpt-5-project"
-    conversation = ConversationManager(llm_client=client)
+    conversation = ConversationManager(llm_client=client, system_prompt=system_prompt)
 
-    def _fake_generate_pnnl(*, max_tokens: int) -> str:
+    def _fake_generate_pnnl(*, max_tokens: int, system_prompt: str) -> str:
         assert max_tokens == DEFAULT_MAX_TOKENS_BY_PROVIDER["pnnl"]
+        assert system_prompt == system_prompt
```
turbomam: I'm not very good at reasoning over what is in the left-hand expression and the right-hand expression when they get compared. Having said that, are you sure this is what you want to do?

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
```diff
@@ -45,3 +40,35 @@
...
+        },
+        {
+            "field_name": "env_local_scale",
+            "reason": "Reason for choosing this field based on the provided information.",
+            "value": ""
+        },
+        {
+            "field_name": "env_medium",
+            "reason": "Reason for choosing this field based on the provided information.",
+            "value": ""
```
turbomam, line 72: Is this a place where you could assert the expected label [CURIE] format?

src/nmdc_metadata_suggestor_ai_tool/recommendation_pipeline.py
```diff
@@ -128,6 +97,22 @@ def run_recommendation_pipeline(
...
     validated_output = clean_and_validate_output(response)
 
+    # get env triad specific recommendations and merge with the general recommendations
+    env_triad_output = get_env_triad_recommendation(
+        context=[submission_context, description, notes],
+        pdf_files=pdf_files,
+        llm_client=llm_client,
+        interface_names=mixs_extensions,
+        max_tokens=max_tokens,
+    )
```
turbomam, line 107: Claude is concerned that one could argue that this is redundant with 

```
response = conversation_manager.generate(max_tokens=max_tokens)
```

in `src/nmdc_metadata_suggestor_ai_tool/recommendation_pipeline.py`

or that at least there should be some comments about why the seconds LLM call is valuable. 

You can dismiss this with no action if you want. I don't feel like I can argue for it.

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
```diff
@@ -45,3 +40,35 @@
 }
 ```
 """
+
+
+env_triad_prompt = """
+Use the provided information to suggest values for the following three metadata fields: env_broad_scale, env_local_scale, and env_medium.
+
+- env_broad_scale: This field should capture the broad environmental category of the sample (e.g., aquatic, terrestrial, host-associated).
```
sierra-moxon:
```suggestion
env_broad_scale: This field should capture the broad environmental category of the sample (e.g., "anthropogenic terrestrial biome [ENVO:01000219]", "freshwater lake biome [ENVO:01000252]", "marine biome ENVO:00000447]").
  ```

File: src/nmdc_metadata_suggestor_ai_tool/system_prompt.py [hunk 8]
```diff
@@ -45,3 +40,35 @@
...
+- env_local_scale: This field should capture the more specific local environment of the sample (e.g., sediment, water column, rhizosphere).
+- env_medium: This field should capture the medium in which the sample was collected (e.g., soil, water, air).
```
sierra-moxon:
```suggestion
env_medium: This field should capture the medium in which the sample was collected ("soil [ENVO:00001998]", "acidic water [ENVO:01000358]", "alluvial paddy field soil [ENVO:00005759]").
```

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
```diff
@@ -45,3 +40,35 @@
 }
 ```
 """
+
+
+env_triad_prompt = """
+Use the provided information to suggest values for the following three metadata fields: env_broad_scale, env_local_scale, and env_medium.
+
+- env_broad_scale: This field should capture the broad environmental category of the sample (e.g., aquatic, terrestrial, host-associated).
+- env_local_scale: This field should capture the more specific local environment of the sample (e.g., sediment, water column, rhizosphere).
```
sierra-moxon:
```suggestion
env_local_scale: This field should capture the more specific local environment of the sample (e.g., "agricultural field [ENVO:00000114]", "sediment [ENVO:00002007]", "rhizosphere [ENVO:00005801]", "aquifer [ENVO:00012408]").
```

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
turbomam: I want to point our that @sierra-moxon 's comment accomplishes two things:

1. clarifies the expected syntax (label [id])
2. shows responses that are better aligned with the [general rules for filling in these MIxS-defined slots with EnvO terms](https://github.com/EnvironmentOntology/envo/wiki/Using-ENVO-with-MIxS)

`env_broad_scale` values should always be a subclass of one of these below. Biomes are strongly preferred. The organism-associated environment annotations are only applicable to the host associated DH interfaces, which may not all be in production yet

@sierra-moxon  or I can help with tools and APIs for validating class membership. We could do it in the NMDC ontology class and ontology relations collection if they have a pure materialized `is_a` closure field

- [EnvO's 'biome'](https://www.ebi.ac.uk/ols4/ontologies/envo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FENVO_00000428)
- [animal-associated environment](https://www.ebi.ac.uk/ols4/ontologies/envo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FENVO_01001002?lang=en)
- [plant-associated environment](https://www.ebi.ac.uk/ols4/ontologies/envo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FENVO_01001001?lang=en)

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
turbomam: subclasses of [environmental material](https://www.ebi.ac.uk/ols4/ontologies/envo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FENVO_00010483)

see also https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/71#discussion_r3086558450

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
turbomam: This is the trickiest slot in the environmental triad

I *really* don't think sediment should go in here! 

`env_local_scale` values should not be biomes or environmental materials. They should be *countable* not *measurable* [astronomical body parts](https://www.ebi.ac.uk/ols4/ontologies/envo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FENVO_01000813?lang=en), minus those other two sets.

See also 

- https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/71#discussion_r3086558450
- https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/71#discussion_r3081849546

src/nmdc_metadata_suggestor_ai_tool/recommendation_pipeline.py
hesspnnl: understandable but keeping this workflow for now.

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
hesspnnl: Im going to make an issue for this!

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
sierra-moxon: agree; thanks, Mark!  

We have a couple of validation modules in the works that can be used here (let's try not to re-write this validation code in this repo, too, if we can help it).  

TL;DR:
For this PR, we might have it easier at the moment: we just need to make sure the term it suggests is from an existing enumeration. If we don't have the interfaces to narrow down which enumeration each suggestion needs to be a part of, then the best we can do with this PR, is confirm if the broad_scale term suggestion is in the (unioned across interfaces) broad_scale "enum".  

When [this goes into nmdc-server](https://github.com/microbiomedata/nmdc-server/pull/2042), by default, if the user accepts all the AI suggestions and then clicks "validate," the UI will run these validation checks (via nmdc-server code) and tell the user what to do (e.g. "you put the same term in env_broad_scale, env_medium, env_local_scale and this violates validation) and won't let the submission be submitted without fixes.  A little awkward maybe for the user as is (e.g. "as a user I don't know how to fill out the env_triad, so I have NMDC's cool AI do it for me.  Nice!  The NMDC AI suggests some plausible terms for me, I accept them, then the NMDC validation says these don't match their rules?")... 

Our tool could be smarter (but I think this is another bolus of work/scope expansion for this PR; we should probably talk about/address these in a separate issue?) in a few different ways: 

1) Pass the suggestions it finds back through the validation steps. If they fail, re-run the suggester with the additional validation context.  Some open questions here: how many times do we do this loop, what if it never validates, there are time tradeoffs here, we're basically starting to write our own agent (the test script we wrote does [this kind of check-before-return-to-the-user round trip but with OLS](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/blob/1fc4762aaa0e904d751f1efa3ad38aef13894ae4/scripts/test_claude_vertex.py#L321) instead of our more robust validation at NMDC).

2) go more agentic.  Give the suggestor tool some Claude skills on how to query the ontology table closures we have, skills on how to validate outside the ontology rules (e.g., regex syntax, required syntax currently checked in data harmonizer code itself, etc.) (how do we run an agent in our stack?  Is it possible?  Alternatively, do we really want to roll our own agent as an alternative?  How does this affect our current user interface? Is an agentic approach more back-and-forth with the user than just a simple "suggestion"?  This feels like a new tool/needs some scoping/brainstorming/discussion.  Maybe there is room in here for some MCPs (but in general, it feels like agents are less productive with MCPs than with just regular old code to run - we need to experiment and do some evals).

sierra-moxon (approved):

Issue #74, opened 2026-04-15
feedback from ai suggester when very little study information is provided


aclum:
It looks to me like the suggester is too aggressive at making suggestions when provided with very limited metadata (I'm not sure if this is realistic use case or not but I feel like this could happen if a user was planning on filling out the study metadata more comprehensively later). I used a study description of 'Bioproducts from Brazil' and it is suggesting fields which don't correspond to the recommended fields for this interface. This relates to #73 in that the interface should be taken into account plus some guidance that is okay for the suggester not to have any suggestions besides the default package recommendations.

<img width="1560" height="774" alt="Image" src="https://github.com/user-attachments/assets/869ce0fc-f1aa-418e-9d72-792354c1f2b1" />

cc @sierra-moxon @hesspnnl

hesspnnl:
I see - for now I can add more instructions to the system prompt and see how that changes the suggestions

sierra-moxon:
@aclum - Identifying the "minimal" amount of information we would need in order to trigger the suggestor to make suggestions was a topic that came up in a couple of different ways in the squad discussions, e.g.: we discussed "what does nmdc-server pass to the suggester, is it individual fields, or the entire submission?" and we also discussed "we need evaluations of models in order to find the LLM that best suggests metadata fields (and values) based on the context we send it" (the context being the number and content of the fields the user sends + the submission schema + any publication text we can retrieve either from the submission itself or fetched separately by the tooling).  

If we could identify a minimum set of fields in a submission that we estimate are sufficient to trigger the suggestor (even if it's not comprehensive or complete), that would be amazing.  Do you have an intuition here that we could build on?

aclum:
I asked over slack what support there was for award DOI lookup, if someone provides that or other DOIs chances are good  the LLM could figure out some reasonable suggestions. I'm also intersted in auto-populating the description if there is an osti doi, no need to make a user enter this twice.
I'd say a publication DOI would be a good indicator, as would a GOLD study ID or NCBI BioProject accession. Website and funding sources doesn't seem specific enough without other more specific information.

You might be able to do some reasoning around the length of the text for Study Description (e.g. less than 250 words is not sufficient)

Issue #55, opened 2026-03-31
Require evidence-grounded justifications, not tautologies


turbomam:
## Problem

The system prompt asks the LLM to output "a short sentence on the reason for choosing each metadata field." In practice, many justifications are tautological — they restate that the field is relevant to the domain without citing anything from the input.

Examples from eval runs against 6 ground truth submissions (gemini-2.5-flash and gpt-4o):

**Tautological (no input evidence):**
- `samp_collec_device`: "The study involves sampling soil, which requires a specific collection device."
- `samp_collec_method`: "Air samples taken, implying a method was used for collection."
- `ph`: "pH is a fundamental soil property."

**Input-grounded (useful):**
- `crop_rotation`: "The abstract explicitly mentions a 'corn-soy-wheat rotation' as part of the agricultural treatments."
- `fire`: "The study investigates the effects of severe wildfire on soil properties."
- `gaseous_environment`: "The study assesses responses to 'exposures to elevated atmospheric CO2 concentrations.'"

The tautological justifications inflate false positives — the model recommends slots simply because they exist in the schema for that package, not because the submission data supports them. In eval, 45% of non-curated recommendations cited the abstract directly, but ~28% were domain-generic or tautological.

## Proposal

Require structured justifications with an explicit evidence category and a text snippet from the input. For example:

```json
{
  "field_name": "crop_rotation",
  "reason_category": "mentioned_in_abstract",
  "evidence_snippet": "corn-soy-wheat rotation as part of the agricultural treatments",
  "reason": "The abstract describes specific crop rotation practices."
}
```

Suggested `reason_category` enum:
- `mentioned_in_abstract` — the abstract or description directly discusses this topic
- `mentioned_in_protocol` — protocol descriptions reference this
- `domain_standard` — standard metadata for this environment/package type (no specific input evidence)
- `experimental_design` — the study design implies this variable was controlled

This would:
1. Let the portal UI show users *why* each field is recommended, with a quote they can verify
2. Let the eval score justification quality separately from slot selection
3. Reduce tautological recommendations — if the model has to produce a snippet and can't find one, it should reconsider recommending the field

## Evidence

From `nmdc-ai-eval` pipeline eval runs (PR https://github.com/microbiomedata/nmdc-ai-eval/pull/35):
- gemini-2.5-flash recommends 17-33 slots per submission; ground truth expects 1-4
- gpt-4o recommends 6-18 slots; higher precision but still many unjustified
[...]

PR #77, opened 2026-04-15
Update system prompt

### LINKED ISSUE #76: update system prompt
When given a simple study description, the tool suggests some logical slots that accurately reference the study description (as expected), but it also suggests some slots by referencing the "AirInterface" schema. If I was a researcher with no prior knowledge, I might get confused if the context I provided (study description) is referenced in the same breadth as a schema I've never seen or been shown before (specifically the third screen shot below)

Image

Image

Image

).

hesspnnl:
This system prompt update targets a few issues - #74, #76, #55 
The updates include tightening the eagerness of suggestions, requiring more explicit and direct reasoning for a field, and ensuring the reasoning does not contain schema specific jargon. **The reasoning will now include a reference to where in the submission the LLM got the info for the slot recommendation**:
'reason': "The submission description focuses on 'marine and terrestrial ecosystems' (submission description)."
'reason': "The submission description focuses on 'marine and terrestrial ecosystems' (abstract)."
'reason': "The submission description focuses on 'marine and terrestrial ecosystems' (notes)."

### DIFF
--- a/src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
@@ -4,17 +4,24 @@ You are an assistant for suggesting metadata for a scientific submission to the 
 - A list of PDF files associated with the publication (if available)
 - Any additional information that may be relevant for suggesting metadata
 
-Use this information to suggest metadata fields for the submission. 
-The metadata fields must only be chosen from the NMDC schema.
-You also need to output a short sentence on the reason for choosing each metadata field, based on the information provided.
+Use this information to suggest metadata fields for the submission. The metadata fields must only be chosen from the NMDC schema.
 
-You should output ONLY THE CHOSEN METADATA FIELDS and the REASON in a JSON list format.
-Do not include any explanations or additional text.
-The metadata fields should be relevant to the content of the abstract and the information provided.
+Important constraints for suggestions and reasons:
+
+REASONING FIELD RULES:
+- Evidence-first: Every `reason` MUST cite a specific piece of the input that supports the suggestion. Prefer an exact short quote (up to 12 words) in quotes, or a concise paraphrase followed by the source label in parentheses (e.g., "abstract", "PDF filename: results.pdf", "additional info").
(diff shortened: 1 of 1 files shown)


src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
```diff
@@ -4,38 +4,34 @@
...
+- No tautology: Do NOT use generic, domain-only justifications (e.g., "pH is a fundamental soil property") unless you also cite supporting input text. Reasons must tie back to the submission input.
+- No schema names: Do not reference  schema packages or technical labels (e.g., "AirInterface"). Reasons should reference only the user's provided input (abstract, filenames, or explicit additional info).
+- Omit weak suggestions: If there is no explicit text and no strong, well-justified inference, DO NOT include the field in the output. Do not add fields solely because they exist in the NMDC schema.
 
 VALUE FIELD RULES:
-- Only fill in the "value" field if you are certain of the specific value for that metadata field based on the provided information.
-- If you are not certain of the specific value for a metadata field, leave the field an empty string ("")
+- Only populate `value` when the input contains the specific value (exact text, identifier, or a resolvable phrase). Otherwise leave `value` as an empty string "".
```
sierra-moxon:
```suggestion
Only populate `value` when the input contains the specific value (exact text). Otherwise, leave `value` as an empty string "".
```

sierra-moxon (approved):
looks good; prompt tweaking is a sensible first step.  if we find ourselves doing a lot of refinement here, we might pause to see if doing some iterative validation and/or tool use (more agentic- whatever we define that as :P ) is more effective.

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
sierra-moxon: I made this change because "identifier" and "resolvable phrase" is_a "exact text", and while I'm not a magician when it comes to prompt engineering, I thought it might be good to not tempt id hallucination by including "identifier" as an example outside of "exact text" (my assumption being that we want an "exact identifier" extracted from the text).

aclum (approved):

Issue #73, opened 2026-04-14
suggester should take interface tab into account


aclum:
I suspect this is because we don't have defined term subsets for the host-associated interface but ideally the suggester would know that a subclass of biome is not appropriate based on an interface of host-associated. 

<img width="1379" height="736" alt="Image" src="https://github.com/user-attachments/assets/d1d6a811-0544-44fe-a127-41b70fd18430" />

cc @sierra-moxon

hesspnnl:
cc @pkalita-lbl this has come up in the meetings before. I can make this an agenda item for next week - but is this something easy to do on the server side? Maybe the LLM output needs another field for "which interface are you getting this slot from?"

PR #78, opened 2026-04-17
Implement structured output

### LINKED ISSUE #75: Implement structured output
There are methods for enforced LLM output, which seems to have great application for the pipelines. 

https://ai.google.dev/gemini-api/docs/structured-output?example=recipe
https://developers.openai.com/api/docs/guides/structured-outputs

hesspnnl:
Utilize provider specific structured output to assist in json response accuracy. 
Question - is this enough validation? should we create some type of looping validation? let the tool just get called again? not worry about it right now?

### DIFF
--- a/src/nmdc_metadata_suggestor_ai_tool/utils/utils.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/utils/utils.py
@@ -7,12 +7,12 @@ from nmdc_metadata_suggestor_ai_tool.models.llm_output import LLMOutput
 logger = logging.getLogger(__name__)
 
 
-def validate_output(structured_output: LLMOutput|str|None) -> LLMOutput:
+def validate_output(structured_output: LLMOutput | str | None) -> LLMOutput:
     """Clean the raw output from the LLM and validate it against the expected schema
     Note
     ----
     GCP and OPENAI providers return the output in different shapes, so this function handles
-    both cases and validates the output against the expected schema. Hence the if statement. 
+    both cases and validates the output against the expected schema. Hence the if statement.
     """
     logger.debug(f"Structured LLM output: {structured_output}")
 
@@ -24,3 +24,4 @@ def validate_output(structured_output: LLMOutput|str|None) -> LLMOutput:
             return validated_output
     except ValidationError as exc:
         raise ValueError(f"LLM response JSON did not match expected output schema: {exc}") from exc
+    return structured_output or LLMOutput()  # return empty output if None or unrecognized format
--- a/src/nmdc_metadata_suggestor_ai_tool/utils/utils.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/utils/utils.py
@@ -1,6 +1,4 @@
-import json
 import logging
-import re
 
 from pydantic import ValidationError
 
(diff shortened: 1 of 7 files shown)


turbomam:
> is this enough validation? should we create some type of looping validation?

Accordin to the [OpenAI docs](https://developers.openai.com/api/docs/guides/structured-outputs):
> Structured Outputs provide reliable type-safety with no need to validate or retry incorrectly formatted responses.

So for the `responses.parse` path I think the answer is no, we don't need a retry layer.

The Gemini path is less clear-cut. I discussed this with Claude. Its read was that a single retry might be worth adding as belt-and-suspenders, but it couldn't find a benchmark proving it's needed when `response_json_schema` enforces the schema server-side. [Instructor](https://python.useinstructor.com/learning/validation/retry_mechanisms/) defaults to up to 2 retries (feeding the `ValidationError` back to the LLM); [LangChain's `RetryWithErrorOutputParser`](https://api.python.langchain.com/en/latest/output_parsers/langchain.output_parsers.retry.RetryWithErrorOutputParser.html) provides the same mechanism without prescribing a count. Neither calibrates against any specific model.

turbomam (commented):
I haven't run this branch locally yet but I really appreciate the fact that you went to the trouble to use the provider-specific solutions for this.

@cmungall's  group has used PydanticAI for the same problem. It bundles the Pydantic-validated structured-output + retry-with-error-feedback loop across providers, so it's worth a look if we end up wanting that layer later. 

You can see the angle I took on the related "LLM returns prose when we wanted parseable output" problem in nmdc-ai-eval #72 and #73. That's the opposite end of the pipeline (the judge, not the generator), but same root cause. My solution there was deleting the LLM judge entirely in favor of deterministic field comparison.

src/nmdc_metadata_suggestor_ai_tool/utils/utils.py
```diff
@@ -9,31 +7,21 @@
...
+        if isinstance(structured_output, LLMOutput):
+            return structured_output
+        elif isinstance(structured_output, str):
+            validated_output = LLMOutput.model_validate_json(structured_output)
+            return validated_output
     except ValidationError as exc:
         raise ValueError(f"LLM response JSON did not match expected output schema: {exc}") from exc
-
-    return validated_output
+    return structured_output or LLMOutput()  # return empty output if None or unrecognized format
```
turbomam, line 27: Could we at least log these conditions where the response gets replaces with `LLMOutput()`?

Does the raised `ValueError` get saved anywhere when this is run headless?

src/nmdc_metadata_suggestor_ai_tool/utils/utils.py
```diff
@@ -9,31 +7,21 @@
...
 
     try:
-        validated_output = LLMOutput.model_validate(parsed_response)
+        if isinstance(structured_output, LLMOutput):
+            return structured_output
+        elif isinstance(structured_output, str):
+            validated_output = LLMOutput.model_validate_json(structured_output)
+            return validated_output
     except ValidationError as exc:
         raise ValueError(f"LLM response JSON did not match expected output schema: {exc}") from exc
```
turbomam, line 26: In the same vein as the older comment below, what is the ideal behavior when we hit this `ValidationError`. I'm not very good with try/except or decisions about skipping, but I think we should have some explicit logging mode

tests/test_recommendation_pipeline.py
```diff
@@ -24,9 +25,13 @@ def __init__(self, response: str) -> None:
         # provide a `.client` with the shape expected by ConversationManager
         # for the PNNL provider: `client.responses.create(...)` returning
         # an object with `output_text`.
+        # Support both `responses.create(...).output_text` and
```
turbomam: I think you completely understand this decision and the consequences, but I don't, and I'm wondering if a future developer would.

I tried reverting `_generate_openai` in `src/nmdc_metadata_suggestor_ai_tool/llm_client.py` back to the pre-PR state:

```python
response = client.responses.create(
    model=self.llm_client.model,
    input=self.messages,
    instructions=system_prompt,
    max_output_tokens=max_tokens,
)
return response.output_text
```

Then ran:

```shell
uv run pytest tests/test_recommendation_pipeline.py -v
```

All 3 tests still passed. The two fast mocked tests can't distinguish `parse` from `create` (the mock returns equivalent JSON either way), and the third (live Gemini) doesn't hit this code path. So if someone later reverted to `create` , pytest wouldn't catch it. Would it be worth an explicit assertion that `parse` was called with `text_format=LLMOutput`?

turbomam (commented):
I have optional ideas about more thorough exception logging and clarifying the roles of `client.responses.create` and `client.responses.parse` but wouldn't argue if you wanted to merge this as is

caufieldjh (approved):
I think this is merge-ready - the logging that @turbomam is talking about is worth returning to and we can iterate on further tweaks, but using the structured output should be a big help in general

src/nmdc_metadata_suggestor_ai_tool/utils/utils.py
hesspnnl: I dont think this would ever happen - the typing was yelling at me... maybe should make a ticket to revisit the typing rules we have

tests/test_recommendation_pipeline.py
hesspnnl: left that in on accident! removed

PR #79, opened 2026-04-20
Add gcp-anthropic access provider for Claude on Vertex

turbomam:
## Summary

- Adds a `gcp-anthropic` access provider parallel to the existing `gcp` (Gemini-only) one
- Uses `AnthropicVertex(credentials=...)` so the same service-account path works for both publishers
- Dispatches via `client.messages.create` (Vertex `rawPredict` / `streamRawPredict`) — which is what Claude-on-Vertex actually requires

## Why

Vertex exposes Gemini and Claude through different endpoints. The existing `access_provider="gcp"` path constructs `google-genai`'s `Client(vertexai=True)` and calls `models.generate_content`, which is Gemini-only. Claude calls through the same path return `400 "not supported in the generateContent API"` even though the `nmdc-llm` project already has the Anthropic publisher enabled (confirmed by Sierra Moxon's 2026-02-06 `vertex-usage` report showing active Claude opus/sonnet/haiku usage).

Empirically verified: with this PR, `claude-haiku-4-5` and `claude-sonnet-4-5@20250929` are reachable on the `nmdc-llm` project through `LLMClient(access_provider="gcp-anthropic")`.

## Notes

- No new env vars — `gcp-anthropic` reads the same `GOOGLE_APPLICATION_CREDENTIALS` / `VERTEX_PROJECT_ID` / `GEMINI_REGION` (or `CLOUD_ML_REGION`) as `gcp`.
- `add_message` emits the Anthropic content-block shape (`{"type": "text", ...}` / `{"type": "document", ...}`). Text-only and text+PDF both supported.
- `gemini_temperature` parameter on `generate()` is ignored for this provider (Claude has its own temperature semantics we can expose separately later if needed).

## Test plan

- [x] Unit tests: default max_tokens dispatch, `add_message` payload shape
- [x] Live test (`test_claude_vertex_generate`, gated by `requires_credentials`) — passes on the `nmdc-llm` project
- [x] `uv run ruff check`, `uv run ruff format --check`, `uv run mypy src/` — all clean
- [ ] Reviewer with GCP creds confirms no regression on existing `gcp` (Gemini) path

Relates to [microbiomedata/nmdc-ai-eval#46](https://github.com/microbiomedata/nmdc-ai-eval/issues/46).


hesspnnl:
Double checked we can use structured outputs and we can. Instructions here https://platform.claude.com/docs/en/build-with-claude/structured-outputs

src/nmdc_metadata_suggestor_ai_tool/llm_client.py
```diff
@@ -249,6 +293,25 @@ def _generate_gcp(
...
+        system_prompt: str,
+    ) -> str:
+        client = cast(AnthropicVertex, self.llm_client.client)
+        message = client.messages.create(
+            model=self.llm_client.model,
+            max_tokens=max_tokens,
+            system=system_prompt,
+            messages=self.messages,
+        )
+        text_parts = [block.text for block in message.content if hasattr(block, "text")]
```
hesspnnl, line 308: what is this line for? In what cases would the response be in blocks for our use case?

README.md
```diff
@@ -27,7 +27,14 @@ Environment variables used by `LLMClient` and `ConversationManager`:
 - `CBORG_KEY`: API key for CBORG (when using `access_provider=cborg`).
 - `CBORG_BASE_URL`: Base URL for the CBORG API.
 
-The `LLMClient` will read the appropriate variables depending on `access_provider` (set to `pnnl`, `cborg`, or `gcp`).
+The `LLMClient` will read the appropriate variables depending on `access_provider`:
+
+- `pnnl` — PNNL AI Incubator (OpenAI-compatible Responses API)
+- `cborg` — CBORG LBNL proxy (OpenAI-compatible)
+- `gcp` — Vertex Gemini via `google-genai` (uses `generateContent`)
+- `gcp-anthropic` — Vertex Claude via `anthropic[vertex]` (uses `rawPredict` / `streamRawPredict`)
```
hesspnnl, line 35: we should not be using streaming with anthropic as the the other providers are not at this time. Please double check that you are not streaming anthropic (does not look like to me) and remove this.

hesspnnl (changes requested):
Overall looks great - thank you for maintaining the backwards compatibility. Wondering if we would want to autodetect the model and keep it to just GCP as the access provider as to me this indicates which credentials the tool is reading as opposed to the llm provider (since cborg and ai incubator are cross llm provider). This may be more of a lift code wise but let me know what you think.

turbomam:
Catching up on this after it sat a while — answers to the three questions:

**Streaming:** not actually streaming. The only "stream" in the file is in the docstring, naming Vertex's REST endpoint (`streamRawPredict` — that's just what Google calls it). The actual code calls `client.messages.create(...)` with no `stream=True`, and reads `message.content` directly rather than iterating a stream. Nothing to remove.

**"What is this line for?"** (`[block.text for block in message.content if hasattr(block, "text")]`) — Anthropic's Messages API returns `content` as a list of blocks, which isn't always just plain text (could include tool-use or thinking blocks depending on config). The `hasattr` filter is there so this doesn't break if a non-text block shows up first — for this use case today it'll always resolve to a single text block, but it's defensive against that changing.

**The bigger one — autodetect the model, keep it to just `gcp`:** agreed, and I've built and tested it. Collapsed `gcp`/`gcp-anthropic` into a single `gcp` provider; the model name picks the Vertex publisher (`claude-*` → AnthropicVertex, anything else → Gemini via google-genai) — the same pattern `pnnl`/`cborg` already use. Small diff, full test suite passes, ruff/mypy clean.

Before I push that: is Claude-via-Vertex for the single-shot pipeline still something you want, given Q4 is pointed at CSV transformation? Component-level Claude-vs-Gemini comparison already works today without this (nmdc-ai-eval's own vertex plugin), so this PR's only remaining value is letting the *full pipeline* run on Claude specifically. Happy to land the design fix + rebase either way, but didn't want to revive a 3-month-stale feature without checking it's still wanted first.

turbomam:
Closing as superseded. Rebasing this onto main surfaced that the agentic architecture (PR #91) already added Claude-via-Vertex access to this same file — the new `agentic()` method uses Claude Agent SDK with GCP auth (`model="claude-sonnet-4-5@20250929"`, a Vertex-style pinned model ID), with the comment "IMPORTANT NOTE: This only works with GCP auth right now." That's this PR's goal, already achieved a different way. No remaining need for a separate `gcp-anthropic` access_provider on LLMClient.

PR #83, opened 2026-04-21
Add automated release

### LINKED ISSUE #82: Add automated release
Per PK:
In many projects I’ve found it convenient to have the creation of a GitHub Release trigger an Action that handles all the necessary building and publishing. It just makes the whole process easier and less error-prone.

Even though we don’t use this package anymore, the geoloc-tools repo has a publish workflow that’s a good baseline: https://github.com/microbiomedata/geoloc-tools/blob/main/.github/workflows/publish.yaml. Note that it doesn’t use a token for PyPI authentication. Instead it uses PyPI’s [Trusted Publishing feature](https://docs.pypi.org/trusted-publishers/).

So, all told I would recommend:

Add a publishing workflow triggered on release creation to the repo.
On PyPI, add at least one or two more people as maintainers of the nmdc-metadata-suggestor-ai-tool package in case you get smashed by a meteor tomorrow.
Also on PyPI, configure GitHub Actions as a trusted publisher

Once all that’s in place, you just have to create the GitHub Release and everything else is automatic.

hesspnnl:
- [x] Add maintainers to PyPi
- [x] Setup Trusted Publisher Management in PyPi
- [x] Add the workflow

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     environment:
       name: pypi
-      url: https://pypi.org/p/nmdc-geoloc-tools
+      url: https://pypi.org/project/nmdc-metadata-suggestor-ai-tool
     permissions:
       # this permission is mandatory for trusted publishing
       id-token: write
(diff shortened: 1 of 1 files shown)


pkalita-lbl (approved):
Looks good!

PR #96, opened 2026-07-07
docs: sync env var docs with what the code reads

turbomam:
Closes #95.

Brings the README, `.env.example`, and the code into agreement. The code reads env vars only through `os.environ`; after this change, every variable it reads is documented and has a placeholder, and nothing it never reads appears in either file.

- Fix `cp .env-example .env` to `.env.example` (the wrong name made the first quickstart step fail).
- Remove `DEFAULT_MODEL`, `MAX_TOKENS`, `TEMPERATURE` from the README Configuration section (no code reads them).
- Remove the commented `OPENALEX_API_KEY` and `UNPAYWALL_EMAIL` from `.env.example` (no code reads them, and there is no Unpaywall code despite the "required" comment). If either is planned, say so and I will restore it.
- Document `CONTACT_EMAIL` and `CLOUD_ML_REGION`.
- Add the ingestion-tuning vars (`NMDC_EDI_MAX_XML_CHARS`, `NMDC_DATAONE_SOLR_MAX_XML_CHARS`, `NMDC_HTTP_RETRY_ATTEMPTS`, `NMDC_HTTP_RETRY_BACKOFF_SECONDS`, `NMDC_HTTP_MAX_RETRY_DELAY_SECONDS`, `NMDC_HTTP_POOL_CONNECTIONS`, `NMDC_HTTP_POOL_MAXSIZE`) to both, with their code defaults.
- Fix a README example that pointed at a non-existent `tests/test_example.py`.

Co-Authored-By: Claude Opus 4.8 (1M context) <noreply@anthropic.com>


README.md
```diff
@@ -23,7 +23,8 @@ Environment variables used by `LLMClient` and `ConversationManager`:
 - `AI_INCUBATOR_BASE_URL`: Base URL for the PNNL AI Incubator API.
 - `GOOGLE_APPLICATION_CREDENTIALS`: Path to a GCP service account JSON file (for Vertex AI).
 - `VERTEX_PROJECT_ID`: (Optional) GCP project id for Vertex. If not provided, the SDK will attempt to infer it from credentials.
-- `GEMINI_REGION`: (Optional) GCP region for Gemini/Vertex (defaults to `us-east5` or `CLOUD_ML_REGION`).
+- `GEMINI_REGION`: (Optional) Vertex region override for Gemini calls (falls back to `CLOUD_ML_REGION`, then `us-east5`).
```
hesspnnl: @caufieldjh @turbomam I did not notice this before, but is there a reason there are so many places to declare gcp region? can we change this to just GCP_REGION ?

hesspnnl (approved):
This looks good - thanks for the initiative on this. I have one concern I commented above. I made some further docs changes that I had been wanting to do.

README.md
caufieldjh: As with all things GCP, the answer is "yes and no". I think the values can be different (https://docs.cloud.google.com/gemini-enterprise-agent-platform/resources/locations kinda gets into this but doesn't fully answer the question).
For our purposes I think it's enough just to set the project ID and region.

hesspnnl: I guess I am asking because in this case they are filling the same value... to me this indicates they are not actually different. At least how we are using them

PR #98, opened 2026-07-07
Add live CBORG provider integration tests

### LINKED ISSUE #61: Add live integration tests of LLM responses from CBORG etc
Will require credentials, so set them up so that CI won't try to run them

turbomam:
Adds opt-in live integration coverage for the CBORG access provider, following the repository's existing provider-test pattern.

## Behavior

- Adds a low-level smoke test that calls the same OpenAI-compatible Responses API endpoint used by the application. This catches authentication, routing, model, and endpoint failures without relying on the separate model-catalog endpoint.
- Adds an end-to-end recommendation-pipeline test that validates non-empty field names and reasons plus the resolved model and access-provider metadata.
- Defaults to the same Gemini model as the production CBORG path while preserving CBORG_TEST_MODEL as an explicit override.
- Caps the integration test at 8,192 output tokens so the test is bounded and does not inherit the provider-wide 128,000-token default.
- Adds credential-aware CBORG pytest fixtures. The tests remain excluded from normal CI and skip when CBORG credentials are absent.

## Rationale

The previous model-catalog smoke check could pass even when the Responses API used by production failed. Live verification demonstrated that exact failure mode: gpt-4o-mini remained advertised while real requests failed because its upstream credits were exhausted. The production-default gemini-2.5-flash completed both the corrected smoke request and the full recommendation fixture.

CBORG requires LBLnet or an authorized IP, so these tests are intentionally not added to the public weekly GitHub Actions workflow.

## Validation

- uv run ruff check
- uv run mypy src
- uv run pytest -q — 257 passed, 59 deselected
- uv run pytest -m integration tests/test_cborg_integration.py -q — 2 passed in 3.05s against live CBORG

## Documentation impact

The test module documents credentials, manual invocation, model override behavior, and why the test is excluded from normal CI. No user-facing application behavior changes.

## Follow-up

Live testing also showed that one provider-wide CBORG output-token default is unsafe across backing models with different context limits. That production policy is intentionally separate in #147.

Closes #61.

### DIFF
--- a/tests/fixtures/llm_auth.py
+++ b/tests/fixtures/llm_auth.py
@@ -1,4 +1,4 @@
-"""Shared pytest fixtures for tests requiring LLM credentials (GCP or PNNL)."""
+"""Shared pytest fixtures for tests requiring LLM credentials (GCP, PNNL, or CBORG)."""
 
 import os
 from pathlib import Path
@@ -38,3 +38,16 @@ def requires_credentials(has_gcp_credentials: bool, has_pnnl_credentials: bool) 
     """Skip the current test when neither GCP nor PNNL credentials are available."""
     if not has_gcp_credentials and not has_pnnl_credentials:
         pytest.skip("Neither GCP nor PNNL credentials available")
+
+
+@pytest.fixture(scope="session")
+def has_cborg_credentials() -> bool:
+    """Return whether CBORG credentials are configured via API key and base URL."""
+    return bool(os.environ.get("CBORG_KEY") and os.environ.get("CBORG_BASE_URL"))
+
+
+@pytest.fixture
+def requires_cborg(has_cborg_credentials: bool) -> None:
+    """Skip the current test when CBORG credentials are not available."""
+    if not has_cborg_credentials:
+        pytest.skip("CBORG credentials (CBORG_KEY, CBORG_BASE_URL) not available")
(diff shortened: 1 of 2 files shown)


turbomam:
This PR is green and mergeable, but it has not yet had a human review and no reviewer is currently requested. Who on the squad is best placed to review the CBORG integration-test assumptions—particularly the explicitly pinned test model and the assertion that `field_name` and `reason` are required while a non-empty suggested `value` is not? This advances #61.

turbomam:
Addressed the suppressed Copilot finding from review 4964022536.

The review was correct: querying /models did not exercise the /responses endpoint used by the application. The smoke test now makes a minimal client.responses.create(...) call instead.

While validating that change, I also updated the test to follow the current production CBORG path:

- The default test model is now the same DEFAULT_GEMINI_MODEL used by LLMClient (currently gemini-2.5-flash); CBORG_TEST_MODEL remains available for explicit overrides.
- The end-to-end test uses an explicit 8,192-token ceiling instead of inheriting the provider-wide 128,000-token default.
- The separate production concern about model-dependent CBORG token limits is tracked in #147.

Validation on current main:

- uv run ruff check — passed
- uv run mypy src — passed
- uv run pytest -q — 257 passed, 59 deselected
- uv run pytest -m integration tests/test_cborg_integration.py -q — 2 passed against live CBORG
- All fresh GitHub checks pass

The branch was also rebased from 140 commits behind main before validation. The PR remains ready for a human reviewer; this automated review does not satisfy that requirement.

hesspnnl:
> test

I would think Harry or Sujay -> lets assign in the meeting today.

caufieldjh (approved):
Thanks @turbomam

PR #128, opened 2026-07-22
Mark real api tests as integration and auto mark to prevent drift

hesspnnl:
Add integration markers and function to prevent integration marking drift.

### DIFF
--- a/tests/test_vertexai_integration.py
+++ b/tests/test_vertexai_integration.py
@@ -27,7 +27,6 @@ def _load_fixture(filename: str) -> dict:
         return json.load(f)
 
 
-@pytest.mark.integration
 @pytest.mark.timeout(INTEGRATION_TIMEOUT)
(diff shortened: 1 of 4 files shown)


tests/test_vertexai_integration.py
turbomam: are you sure this needs to be asserted for each function in the module?

turbomam: another potential duplicate of a module-level assertion?

turbomam: especially since all functions in this file have the same decorators?

turbomam (approved):
This is good. I only have one suggestion: multiple identical decorators on functions in one source file could probably be removed. Depending on your outlook that could make the readability better or worse!

I also added an issue about determining which integration tests should be run in CI

- https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/134

PR #132, opened 2026-07-24
Skills and functions for supplement retrieval

caufieldjh:
Closes #117.

## Summary

Adds an Agent Skill (`supplement-retrieval`) and helper functions to retrieve **supplementary materials for a manuscript (or dataset) DOI**, scoped to what helps the NMDC submission system characterize submissions and samples.

Design follows the issue's intent to **value the useful supplement types, and favor performance over comprehensiveness**:

- **Value filter:** only high-value kinds are kept by default — *tabular* (`csv/tsv/xlsx/…`, per-sample metadata tables) and *documents* (`pdf/docx/txt/…`, supplementary methods). Images, media, sequence data, and archives are classified and skipped. Captions from JATS/repository metadata let files be judged by content, not just extension.
- **Performance-first:** single attempt per source, streamed size-bounded downloads, oversized files skipped by reported size *before* reading, zip/tar path-traversal and DOCTYPE/ENTITY guards, and a **shared download budget** across sources so the total never exceeds the global caps.

## DOI contract (layered resolution)

`retrieve_supplements(doi, text=None, follow_related=True)` routes by DOI type:

- **Data-repository DOI** (`10.5061` Dryad, `10.5281` Zenodo, `10.6084` Figshare) → fetch that repo's files.
- **Publication DOI** → hosted supplements (Europe PMC → NCBI PMC OA fallback) **merged** with linked data-repository datasets found via Crossref/DataCite relation metadata (`follow_related`) or mined from `text`. Sequence/proteomics accessions (PRJNA…/SRR…/GSE…) found in `text` are surfaced in `detected_accessions` but not retrieved.

## Sources

1. **Europe PMC** `supplementaryFiles` ZIP (open-access), captions from JATS `fullTextXML`.
2. **NCBI PMC OA** `.tar.gz` package — fallback when Europe PMC has none; captions from the bundled `.nxml`.
3. **Dryad / Zenodo / Figshare** — data-repository file retrieval. Dryad's own download API is auth-gated (401) and Cloudflare-gated, so Dryad content is fetched from its **Zenodo mirror** (degrades gracefully when a dataset isn't mirrored).

## Caveats

- Text mining is precision-over-recall; sequence accessions are reported, not fetched.
- Dryad↔Zenodo mirroring wound down ~2020–2022, so the newest Dryad datasets may not be retrievable (returns a download-gated listing).
- Reusable file-type taxonomy lives in a new `file_kinds` module, not `constants.py`.

## Testing

- 35 hermetic `responses`-mocked unit tests; ruff + mypy clean.
- Opt-in `integration`-marked live smoke tests verified against real Zenodo/Figshare/Dryad APIs.

🤖 Generated with [Claude Code](https://claude.com/claude-code), mostly


.claude/skills/supplement-retrieval/SKILL.md
```diff
@@ -0,0 +1,93 @@
+---
+name: supplement-retrieval
+description: Use this skill to retrieve supplementary materials for a manuscript DOI, prioritizing the file types useful for characterizing NMDC submissions and samples (tabular metadata and supplementary documents). Performance-first: single attempt, no repeated retries.
```
hesspnnl: Extra colon causing markdown syntax errors

```suggestion
description: Use this skill to retrieve supplementary materials for a manuscript DOI, prioritizing the file types useful for characterizing NMDC submissions and samples (tabular metadata and supplementary documents). Performance-first, single attempt, no repeated retries.
```

.claude/skills/supplement-retrieval/SKILL.md
```diff
@@ -0,0 +1,93 @@
...
+
+Pass `sources=[...]` to run an explicit list (`"europepmc"`, `"pmc_oa"`, `"dryad"`,
+`"zenodo"`, `"figshare"`), or `follow_related=False` to skip relation-following.
+To cheaply check availability *before* downloading, call
+`find_supplement_source_europepmc(doi)` or `find_supplement_source_pmc_oa(pmcid)`.
+
+For the full API (per-source helpers, caps, kind filtering, temp-file cleanup),
+see [refs/programmatic.md](refs/programmatic.md).
+
+## Agentic fallback (optional, one shot)
```
caufieldjh: @hesspnnl mentioned this at the call on Aug 4 - this broader web search may be out of scope (and agent permissions)

hesspnnl: I am curious in testing if you have seen it use this or need it?

src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/retrieve_supplements.py
hesspnnl, line 1: I think this would be more digestible following a more deconstructed pattern (like how we have the `doi` directory). Something like:
```bash
publication_ingestion/
  supplements/
    __init__.py          # re-exports retrieve_supplements (public API)
    shared.py           # _Member, _select_members, _materialize_member, SupplementCaps, JATS parsing
    europepmc.py        # retrieve_supplements_from_europepmc & find/fetch helpers
    pmc_oa.py           # retrieve_supplements_from_pmc_oa
    dryad.py            # retrieve_supplements_from_dryad
    zenodo.py           # retrieve_supplements_from_zenodo
    figshare.py         # retrieve_supplements_from_figshare
    related_dois.py     # find_related_data_dois, extract_dataset_dois_from_text, extract_accessions_from_text
    retrieve.py
```

src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/retrieve_supplements.py
```diff
@@ -0,0 +1,1416 @@
...
+    useful_kinds: frozenset[SupplementKind]
+    max_files: int
+    max_file_bytes: int
+    max_total_bytes: int
+    max_archive_bytes: int
+    max_text_chars: int
+    save_dir: str | None
+
+
+def _resolve_caps(
```
hesspnnl: I dont see the value in this function - i think it should be deleted and replaced with direct construction

src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/retrieve_supplements.py
```diff
@@ -0,0 +1,1416 @@
...
+                },
+            )  # type: ignore[arg-type]
+        )
+
+    merged = _merge_results(doi, results, accessions, max_files, max_total_bytes, save_dir=save_dir)
+    merged.pmcid = pmcid or merged.pmcid
+    return merged
+
+
+def _run_source(
```
hesspnnl: claude response - _run_source makes a redundant API call for pmc_oa line 1327. when pmcid=None, it calls find_supplement_source_europepmc again to resolve it, even though the normal orchestrator path at line 1276 already did this. The explicit sources path (sources=["pmc_oa"]) hits the EuropePMC search API twice. Worth passing the resolved pmcid into _run_source rather than re-fetching it.

src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/retrieve_supplements.py
```diff
@@ -0,0 +1,1416 @@
...
+    attempts: list[str] = []
+    for res in results:
+        attempts.extend(res.attempts)
+
+    kept: list[SupplementFile] = []
+    seen: set[tuple[str | None, str]] = set()
+    total_bytes = 0
+    for res in results:
+        for file in res.files:
+            key = (file.source, os.path.basename(file.filename))
```
hesspnnl: claude comment:
Dedup key in _merge_results uses (source, basename), this would over-merge two different files from the same source that happen to share a basename (e.g. Table_S1.xlsx from both a PMC ZIP and a linked Dryad deposit). Keying on (source, full filename) would be more precise.

hesspnnl (changes requested):
Some comments down below as well. 

- the `_` prefixes are standard Python but at this density they stop conveying meaning to me. When this gets split into a package, id prefer most of them to drop.
- Add a test that targets the agent using this skill

hesspnnl:
I will also request Mark to review, since he has closer subject matter expertise

turbomam:
I am very impressed with this PR but I will have several suggestions in over the next ~~ten minutes~~ hour or so

.claude/skills/supplement-retrieval/SKILL.md
caufieldjh: I haven't seen it used successfully yet. The emphasis on "one" web_fetch attempt may not be helping. I'm going to remove this section.

src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/retrieve_supplements.py
caufieldjh: I see what you mean - will try a restructure like that

caufieldjh:
One thing I only just learned: the existing PMC OA web service API/FTP is being deprecated. That is, it's intended to be fully deprecated later this month (see https://ncbiinsights.ncbi.nlm.nih.gov/2026/02/12/pmc-article-dataset-distribution-services/).
It's intended to be on a new AWS Open Data service, but in lieu of setting that up I'm just going to remove the PMC OA details.

turbomam (approved):
Everything I was concerned about yesterday has already been fixed! @hesspnnl and @caufieldjh, you are a good team!

I ran the tests, including the integration tests, against https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/commit/c1f494c7c2eff0a89891b0633d351802d13e96e7.

I also called `retrieve_supplements` directly:

```python
from nmdc_metadata_suggestor_ai_tool.publication_ingestion.supplements import retrieve_supplements
r = retrieve_supplements("10.1038/s41564-020-00861-0", follow_related=False)
r.source, r.error, len(r.files)
[f.filename for f in r.files]
```

That returned 10 files with `error=None` today. Yesterday, against https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/commit/899e83bd5813f33b9826eb1ee1f1499242748f2f, the same call returned 0 files and a 404 on `https://www.ebi.ac.uk/europepmc/webservices/rest/MED/33526884/supplementaryFiles`.

The two Europe PMC URLs were wrong, but every test passed, because the mocked tests build their URLs from the same constants the code uses. Your `test_europepmc_live_url_is_reachable` is the test that catches that. I ran it against both commits and it fails on the older one and passes on the current one.

I approve this, but I opened this as potential followup on the prioritization of the downlaods
- https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/142

src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/supplements/zenodo.py
```diff
@@ -0,0 +1,110 @@
...
+
+def zenodo_file_members(doi: str, max_file_bytes: int) -> tuple[list[Member], str | None]:
+    """Return download members for the Zenodo record matching *doi*.
+
+    Shared by the Zenodo retriever and the Dryad retriever (Dryad datasets are
+    mirrored on Zenodo, and Dryad's own download API is auth-gated). Returns
+    ``(members, error)`` where ``error`` is set when no record/files were found.
+    """
+    try:
+        response = request_with_retry(
```
hesspnnl: Non blocking suggestion:
`zenodo_file_members` in `src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/supplements/zenodo.py` duplicates logic from `src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/zenodo.py`

You could expose the raw response in supplements/ and import in doi_ingestion so that it can call a shared query helper instead of reimplementation
Might be more effort than it is worth - but food for thought on cutting down line contribution!

src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/supplements/shared.py
```diff
@@ -0,0 +1,369 @@
...
+
+
+def caption_key(filename: str) -> str:
+    """Normalize a filename/href to a match key (basename, no extension, lower)."""
+    base = os.path.basename(filename).lower()
+    stem, _, ext = base.rpartition(".")
+    return stem if (ext and stem) else base
+
+
+def parse_supplement_captions(xml_text: str | bytes) -> dict[str, str]:
```
hesspnnl, line 169: non-blocking suggestion:
The XML defense here looks similar to the one in `doi_ingestion/edi.py`. Now there are two implementations of the same security pattern. Could be worth sharing so there is one place to update in the future if needed

src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/supplements/figshare.py
```diff
@@ -0,0 +1,115 @@
...
+                size=int(file_entry.get("size") or 0),
+                read=url_reader(url, caps.max_file_bytes),
+            )
+        )
+
+    apply_selection(result, members, caps, None, "Figshare record contained no high-value files")
+    return result
+
+
+def figshare_detail_for_doi(doi: str) -> dict[str, Any] | None:
```
hesspnnl, line 80: Both doi_ingestion/figshare.py and publication_ingestion/supplements/figshare.py hit the same two Figshare endpoints (articles API, then collections API) to resolve a DOI to a record, and both use a similar "try articles first, fall back to collections" pattern. The supplements side extracted figshare_detail_for_doi as a named function; doi_ingestion appears to inline the same two-step lookup. If figshare_detail_for_doi were shared, doi_ingestion could import and reuse it directly and reduce code lines, and make it easier to maintain in the future

.claude/skills/supplement-retrieval/refs/programmatic.md
```diff
@@ -0,0 +1,160 @@
...
+    retrieve_supplements_from_zenodo,  # Zenodo record files (by DOI/concept DOI)
+    retrieve_supplements_from_figshare,  # Figshare article/collection DOI files
+    find_related_data_dois,  # publication DOI -> linked data-repo DOIs
+    extract_dataset_dois_from_text,  # text -> Dryad/Zenodo/Figshare DOIs
+    extract_accessions_from_text,  # text -> PRJNA…/SRR…/GSE… accessions
+    find_supplement_source_europepmc,  # availability check, no download
+    parse_supplement_captions,  # JATS XML -> {filename_key: caption}
+    classify_supplement,  # filename -> SupplementKind
+    is_dryad_doi,  # True for 10.5061/… DOIs
+    DEFAULT_USEFUL_KINDS,  # {TABULAR, DOCUMENT}
```
hesspnnl, line 27: looks like this is imported from the wrong file. should import from -`nmdc_metadata_suggestor_ai_tool.file_kinds` instead of `nmdc_metadata_suggestor_ai_tool.publication_ingestion.supplements`

hesspnnl (changes requested):
This looks great Harry! I was able to go through and find some small areas of duplication that potentially could be annoying maintenance tasks in the future. Once these are handled I am happy to merge in.

hesspnnl (approved):
Awesome - thanks for making those chages!

PR #139, opened 2026-08-06
Create Skill to enforce DOI copywrite for AI use

### LINKED ISSUE #111: Create new skill: DOI copywrite check
This skill will check copy write of DOIs and materials to determine if they can be ingested. 

Utilize librarian guidelines.

hesspnnl:
Create a Skill that helps the agent avoid using DOIs that are not cleared for AI use. As a side affect of this, it will also create clear documentation around current copywrite and AI policies. 
IMPORTANT - This PR also makes changes to the agentic path:
- adds a permission override for bash commands and tool use. I was noticing that the agent is really held back by not having permissions for this. I think it should be opened up or else we are **not** leveraging agentic processes well. As a result - **we will need** to containerize this when we deploy to nmdc-sever or we are really opening ourselves up to security risks. 
- Updates the **orchestration prompt** to discourage WebSearch and any bash commands that are outside of direct skill usage. 
- Updates the **langfuse tracing** to log each agent message - this will increase logs and observations but I was finding I wanted more info from the logs than I was getting.

### DIFF
--- a/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
@@ -381,7 +381,7 @@ class ConversationManager:
             model=model,
             system_prompt=orchestrator_prompt,
             output_format={"type": "json_schema", "schema": LLMOutput.model_json_schema()},
-            permission_mode="bypassPermissions",
+            permission_mode=cast(Any, AGENT_PERMISSION_MODE),
         )
 
         if message is None:
--- a/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
@@ -358,7 +358,11 @@ class ConversationManager:
         If None, starts a new session.
 
         """
-        model = DEFAULT_CLAUDE_MODEL if self.llm_client.access_provider == "gcp" else self.llm_client.model
+        model = (
+            DEFAULT_CLAUDE_MODEL
+            if self.llm_client.access_provider == "gcp"
+            else self.llm_client.model
+        )
         # set env variable to enable Claude Agent SDK to pick up GCP credentials
         # Claude Agent SDK requires a Claude model, not a Gemini model, even on Vertex AI
         options = ClaudeAgentOptions(
--- a/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
@@ -23,6 +23,7 @@ from nmdc_metadata_suggestor_ai_tool.models.llm_output import LLMOutput
 from nmdc_metadata_suggestor_ai_tool.system_prompt import orchestrator_prompt
 from nmdc_metadata_suggestor_ai_tool.tracing import (
     langfuse_client,
+    log_assistant_message,
     observe,
     propagate_attributes,
 )
@@ -394,11 +395,7 @@ class ConversationManager:
                     session_id = event.data["session_id"]
                 elif isinstance(event, AssistantMessage):
                     print(f"Assistant: {event.content}")
-                    if langfuse_client is not None:
-                        langfuse_client.create_event(
-                            name="assistant_message",
-                            output=event.content,
-                        )
+                    log_assistant_message(event.content)
                 elif isinstance(event, ResultMessage):
                     result = self.unwrap_structured_output(event.structured_output)
             if langfuse_client is not None:
@@ -419,11 +416,7 @@ class ConversationManager:
                         session_id = event.data["session_id"]
                     elif isinstance(event, AssistantMessage):
                         print(f"Assistant: {event.content}")
-                        if langfuse_client is not None:
-                            langfuse_client.create_event(
-                                name="assistant_message",
-                                output=event.content,
-                            )
+                        log_assistant_message(event.content)
                     elif isinstance(event, ResultMessage):
                         result = self.unwrap_structured_output(event.structured_output)
             if langfuse_client is not None:
--- a/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
@@ -366,6 +366,7 @@ class ConversationManager:
             model=model,
             system_prompt=orchestrator_prompt,
             output_format={"type": "json_schema", "schema": LLMOutput.model_json_schema()},
+            permission_mode="bypassPermissions",
         )
 
         if message is None:
@@ -389,6 +390,11 @@ class ConversationManager:
                     session_id = event.data["session_id"]
                 elif isinstance(event, AssistantMessage):
                     print(f"Assistant: {event.content}")
+                    if langfuse_client is not None:
+                        langfuse_client.create_event(
+                            name="assistant_message",
+                            output=event.content,
+                        )
                 elif isinstance(event, ResultMessage):
                     result = self.unwrap_structured_output(event.structured_output)
             if langfuse_client is not None:
@@ -407,6 +413,13 @@ class ConversationManager:
                 ):
                     if isinstance(event, SystemMessage) and event.subtype == "init":
                         session_id = event.data["session_id"]
+                    elif isinstance(event, AssistantMessage):
+                        print(f"Assistant: {event.content}")
+                        if langfuse_client is not None:
+                            langfuse_client.create_event(
+                                name="assistant_message",
+                                output=event.content,
+                            )
                     elif isinstance(event, ResultMessage):
                         result = self.unwrap_structured_output(event.structured_output)
             if langfuse_client is not None:
--- a/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/llm_client.py
@@ -32,6 +32,16 @@ load_dotenv()
 
 DEFAULT_GCP_REGION = "us-east5"
 
+# Set CLAUDE_AGENT_BYPASS_PERMISSIONS=true to disable the SDK's permission guardrails.
+# Disabling allows Claude to run bash commands and WebSearch,
+# which improves preformance but increases security risks.
+# Defaults to "default" which requires explicit permissions for shell commands and WebSearch.
+AGENT_PERMISSION_MODE = (
+    "bypassPermissions"
+    if os.environ.get("CLAUDE_AGENT_BYPASS_PERMISSIONS", "") == "true"
+    else "default"
+)
+
 
 GEMINI_MODELS = [
     "gemini-2.5-pro",
--- a/src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
@@ -43,6 +43,8 @@ orchestrator_prompt = """
 Use the nmdc-metadata-suggestor skill to process the input and return a merged LLMOutput JSON.
 
 CRITICAL EXECUTION RULES:
+- Do not run shell commands, execute code, or read/write files outside of the skills you have been given. Only use tools as directed by skill instructions.
+- Do not run WebSearch or any external API calls. Only use the skills provided.
 - Do NOT call StructuredOutput until ALL pipeline steps are fully complete.
 - Call StructuredOutput exactly once — as the very last action — with the final LLMOutput JSON (metadata_fields list).
 - If the stop hook fires mid-pipeline, ignore it and continue executing the remaining steps before calling StructuredOutput.
--- a/.claude/skills/build-study-context/SKILL.md
+++ b/.claude/skills/build-study-context/SKILL.md
@@ -46,8 +46,11 @@ result = get_doi_description_or_abstract(
 )
 ```
 
-- Add `result.context` to the evidence texts (label it with the DOI value)
-- Accumulate all `result.publication_urls` into a list
+Then apply the **doi-copyright-check** skill to the returned metadata:
+
+- If `verdict == "allowed"`: add `result.context` to the evidence texts (label it with the DOI value)
+- If `verdict == "not_allowed"` or `"uncertain"`: **exclude** the content; add the DOI to a `skipped_dois` list and log a warning
+- Accumulate all `result.publication_urls` into a list regardless of verdict (URLs themselves are not restricted content)
 
 ### 3 — Fetch PDF content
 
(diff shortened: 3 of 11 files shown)


hesspnnl:
Review direction:
@sujaypatil96 can you take a look at the changes to llm_client.py? Specifically testing the added langfuse logging to see if you like it. it is more verbose.
@caufieldjh can you read over the new skill and check for validity? i tested with a few DOIs, but if you have specific ones where you are like "this shouldn't be used" that would be great to test with. 
@turbomam Harry and Sujay, thoughts on opening the permissions? I created a follow on task for containerizing this when we deploy to server, but there are implications of allowing everything as a blanket.

src/nmdc_metadata_suggestor_ai_tool/llm_client.py
```diff
@@ -407,6 +417,13 @@ async def agentic(
                 ):
                     if isinstance(event, SystemMessage) and event.subtype == "init":
                         session_id = event.data["session_id"]
+                    elif isinstance(event, AssistantMessage):
+                        print(f"Assistant: {event.content}")
+                        if langfuse_client is not None:
+                            langfuse_client.create_event(
+                                name="assistant_message",
+                                output=event.content,
+                            )
```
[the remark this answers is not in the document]
hesspnnl: we could make this a specific function in tracing. good idea.

src/nmdc_metadata_suggestor_ai_tool/llm_client.py
```diff
@@ -389,6 +394,11 @@ async def agentic(
                     session_id = event.data["session_id"]
                 elif isinstance(event, AssistantMessage):
                     print(f"Assistant: {event.content}")
+                    if langfuse_client is not None:
+                        langfuse_client.create_event(
+                            name="assistant_message",
+                            output=event.content,
+                        )
```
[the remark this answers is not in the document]
hesspnnl: I am aware. We can cut down later as we find issues.

src/nmdc_metadata_suggestor_ai_tool/llm_client.py
```diff
@@ -358,14 +358,19 @@ async def agentic(
...
+        )
         # set env variable to enable Claude Agent SDK to pick up GCP credentials
         # Claude Agent SDK requires a Claude model, not a Gemini model, even on Vertex AI
         options = ClaudeAgentOptions(
             skills="all",
             model=model,
             system_prompt=orchestrator_prompt,
             output_format={"type": "json_schema", "schema": LLMOutput.model_json_schema()},
+            permission_mode="bypassPermissions",
         )
```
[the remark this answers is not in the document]
hesspnnl: I am aware - this is a concern.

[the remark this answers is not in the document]
hesspnnl: added a config option for now - will move to a container version soon

src/nmdc_metadata_suggestor_ai_tool/llm_client.py
```diff
@@ -31,6 +32,16 @@
...
 
+# Set CLAUDE_AGENT_BYPASS_PERMISSIONS=true to disable the SDK's permission guardrails.
+# Disabling allows Claude to run bash commands and WebSearch,
+# which improves preformance but increases security risks.
+# Defaults to "default" which requires explicit permissions for shell commands and WebSearch.
+AGENT_PERMISSION_MODE = (
+    "bypassPermissions"
+    if os.environ.get("CLAUDE_AGENT_BYPASS_PERMISSIONS", "") == "true"
+    else "default"
+)
```
sujaypatil96, line 43: Unrelated to what you asked me to review - but in https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144 i've added a `.claude/settings.json` file with an allowlist for specific bash commands. I'm wondering if we should maintain a list like that here too and that we avoid some of the _security risks_ (that you've mentioned in your PR description) that we are worried about too.

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
```diff
@@ -43,6 +43,8 @@
 Use the nmdc-metadata-suggestor skill to process the input and return a merged LLMOutput JSON.
 
 CRITICAL EXECUTION RULES:
+- Do not run shell commands, execute code, or read/write files outside of the skills you have been given. Only use tools as directed by skill instructions.
+- Do not run WebSearch or any external API calls. Only use the skills provided.
```
sujaypatil96, line 47: Can we look at a trace for an example DOI that you tested with (either one of the fixtures in this PR) to make sure that both of these rules are followed? I see the agent making API calls to the Crossref API in one of my agent traces.

sujaypatil96, line 47: Claude recommendation:

> Plumb message['license'] through crossref.py, call check_doi_copyright() from build-study-context instead of leaving it dead, and the check works under the existing

sujaypatil96 (commented):
Left a few comments on this PR.

JSON from trace for open access publication in this PR:

```json
{
    "$PARAMETER_NAME": {
        "metadata_fields": [
            {
                "field_name": "store_cond",
                "value": "frozen -80C",
                "reason": "'Samples stored at -80C prior to DNA extraction' (submission notes).",
                "id": null
            }
        ],
        "model": "claude-sonnet-4-20250514",
        "access_provider": "anthropic"
    }
}
```

- [x] Fabricated/LLM-populated `LLMOutput` fields
- [x] The payload above looks good but want to point out the manufactured top-level placeholder for `metadata_fields` (reference to `StructuredOutput` argument) -- fix for this in https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/ (see `structured_output_from_tool_use()`)

Agent trace for the restricted publication: [agent_trace_restricted_pub_bypass.json](https://github.com/user-attachments/files/31706394/agent_trace_restricted_pub_bypass.json)

turbomam (commented):
Claude summary of my analyses:

**Open them, but not as a blanket.** The case for opening is measured, not hypothetical: `docs/agent-permissions.md` on https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144 records traces where 95 of 104 observations were permission denials, including a 100-sample run where the agent tried twice to reach the ontology, was denied both times, and still produced all 300 suggestions. The denials did not fail loudly. The agent fell back to the value sets already in its prompt and returned confident output with a whole tier of the product silently not running. So the closed default is not the safe option it looks like, and your instinct that the agent is being held back is backed by data.

The blanket is the part I would not do. `bypassPermissions` removes the enforcement mechanism entirely, and this PR itself shows what is left behind when it goes: the new prompt lines say "Do not run shell commands... Do not run WebSearch or any external API calls", and @sujaypatil96 reports already seeing Crossref API calls in his agent traces (https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/139#discussion_r3907157973). A prompt sentence is a request. The allowlist he added on https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144 (`.claude/agent-settings.json`, denying `curl`, `wget`, `ssh`, `git push`, package installs and similar) opens everything the agent legitimately needs while keeping the rules in configuration the model cannot ignore. You said on that PR today it seems more elegant, and I agree. Since you offered in the squad channel to split the scope creep out anyway, my suggestion is: land the Digital Object Identifier (DOI) skill from this PR, and make the permissions change an adoption of 144's settings file rather than a bypass toggle.

**On the container as the follow-on (https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/148, Containerize our agent):** it fixes less than it appears to. A container bounds what the agent can reach on the host. It does not bound the credentials we put into the agent's own environment: `.env.example` on this branch carries `LANGFUSE_SECRET_KEY`, `AI_INCUBATOR_KEY`, `CBORG_KEY` and `GOOGLE_APPLICATION_CREDENTIALS`, and an agent with shell and network access has what it needs to read an environment variable and send it somewhere. The deny-list addresses that directly; a container does not. Worth settling before 148 is scoped.

---

Two findings from kicking the tires, offered to the validity review rather than as my lane. The decision rules themselves look right to me, and the tests are real: 23 pass, and forcing wrong verdicts makes them fail in exactly the directions they should.

**The license never reaches the check.** The skill says to look first in "a `license` or `rights` field in the DOI source metadata", and Crossref does return one; `api.crossref.org/works/<doi>` carries a `license` array with the Creative Commons URL. But `crossref.py` on this branch never reads that field, and `ResolverContext`, the object handed back to the agent, has no license field. So the lookup the skill describes cannot succeed, and the agent is left scanning abstract text for phrases like "licensed under a Creative Commons", which most abstracts omit even when the paper is CC BY. The everyday outcome will be "license unclear", which the rules correctly map to exclude, so the practical effect is dropping usable papers rather than blocking restricted ones. @sujaypatil96's suggestion at https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/139#discussion_r3907180172 is the fix, plus one step: have build-study-context call `check_doi_copyright()` with the plumbed license. Today that function is invoked only by its own tests, and the skill restates the same rules in prose, so the tested implementation and the running one are different things.

**Two no-license documents get opposite verdicts.** Against live Crossref metadata: a Science article (no license, no abstract in Crossref) comes back `uncertain`, while a Nucleic Acids Research article (no license, abstract present) comes back `not_allowed`, because the abstract scan runs, finds no CC phrase, and treats that as a negative finding. Absence of a CC phrase in an abstract says nothing about the license, and that Nucleic Acids Research article is Oxford and very likely CC BY. Suggested fix: the abstract scan should only ever return "allowed" or "no signal", never "not_allowed", so a missing license reaches `uncertain` by the same path whether or not an abstract happens to exist. Reproduce without credentials: `curl -s https://api.crossref.org/works/10.1093/nar/gkaa1105 | jq .message.license`, same for `10.1126/science.aav2566`, then hand each license and abstract to `check_doi_copyright`.

The per-message Langfuse logging looks right, and matches what I have wanted from the traces too.

src/nmdc_metadata_suggestor_ai_tool/llm_client.py
hesspnnl: as discussed in 9/1 meeting i remove the bypass from this pr in favor of the settings profile

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
hesspnnl: im not sure i follow this suggestion?

hesspnnl: yep!
[doi allowed](https://us.cloud.langfuse.com/project/cmrjig05g03kgad0d67phwxds/traces/3e8850a6870cf77e75ab1af2c4614b08?observation=c6e4090a6dcdcb8a&timestamp=2026-09-08T19:33:42.246Z&traceId=3e8850a6870cf77e75ab1af2c4614b08)
doi not allowed

hesspnnl:
@sujaypatil96 check out the changes commit 726b5139f15c7edc12cd87c3dc63571a575b2466 introduces.

The old approach stored allow/deny rules in .claude/settings.json and .claude/agent-settings.json, but those files only take effect if you've clicked through Claude Code's workspace trust dialog — which you can't do on a server or a fresh teammate setup. This made permissions silently do nothing in headless runs.
Instead, the same rules now live in agent_permissions.py as plain Python constants and a PreToolUse hook registered directly on ClaudeAgentOptions. No config files, no trust dialog, works the same everywhere.
One gotcha along the way: the SDK has a can_use_tool callback that seems like the right fit, but it requires streaming mode, which caused the agent to hang with no output. The PreToolUse hook does the same job without that constraint.

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
sujaypatil96: Right, yes, sorry, looks like I copied the suggestion the suggestion over only partially (by mistake).

The recommendation came from Claude after it looked at a few agent traces and saw that in the `bypassPermissions` runs ([pr139-cc-by-bypass](https://us.cloud.langfuse.com/project/cmrjig05g03kgad0d67phwxds/traces?filter=type%3BstringOptions%3B%3Bnone+of%3BTOOL%257CEVENT%257CAGENT%2Cname%3BstringOptions%3B%3Bnone+of%3Bagentic%257Cenv-triad-fixture-run&peek=c1a58d31ee320ba0&observation=c1a58d31ee320ba0&traceId=3e7d977c0fc27e173d6525cad9f3f2da&timestamp=2026-09-01T18%3A09%3A17.389Z) / [pr139-restricted-bypass](https://us.cloud.langfuse.com/project/cmrjig05g03kgad0d67phwxds/traces?filter=type%3BstringOptions%3B%3Bnone+of%3BTOOL%257CEVENT%257CAGENT%2Cname%3BstringOptions%3B%3Bnone+of%3Bagentic%257Cenv-triad-fixture-run&peek=0492fb6d74d495d5&observation=0492fb6d74d495d5&traceId=b64dda3d3318da129ea937e7d77eb820&timestamp=2026-09-01T18%3A24%3A03.759Z)), the only way the agent obtained a license was by hand-writing its own call:

```bash
httpx.get('https://api.crossref.org/works/10.1038/s41564-020-00861-0')
license_info = data['message'].get('license', [])
```

The motivation for this suggestion was the fact that `crossref.py` has no _license_ / _rights_ handling, `SourceRetrievalResult` doesn't have a license field, and check_doi_copyright still has zero callers outside its own module and tests.

The rule that i've highlighted on this review comment says "Do not run WebSearch or any external API calls" but it looks like a hand-rolled call by the agent is the only way for it to obtain a license.

sujaypatil96:
@hesspnnl awesome, I love the hooks based policy you've set up in the `agent_permissions.py` file. I see you've moved the agent settings permissions profiles (`claude/settings.json` / `claude/agent-settings.json`) with tool allow/deny lists in there.

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
sujaypatil96: Thanks for sharing these!

hesspnnl: i see!! thanks for pointing this out. i committed fix for this in 811206ed9df23d7380423ce1a651205f717bd7e5

sujaypatil96 (approved):
Thanks for the recent commits and all the work on this PR @hesspnnl - all the comments I raised above have been addressed 🚀 this PR is ready to be merged!

PR #91, opened 2026-05-26
Agentic Architecture

hesspnnl:
Adding skills and an agentic loop

PR #144, opened 2026-08-11
ENVO term expansion and validation for the env-triad skill

### LINKED ISSUE #109: Create new skill: ENVO term expansion
Expand the env triad skill to include more sources. This will assist in more detailed ENVO IDs and can include IDs we do not currently have in the submission enums. 

First use the submission subset, if there's nothing that seems appropriate look at envo, follow the train down....

Outstanding questions:

- What outside sources should we reach out to? Do APIs exist? Do we need to create new tools that reach out to APIs? We need to be conscious about just pointing agent to a URL - opens agent to security risks. 

Deliverable:

Enhanced env-triad skill that recommends ENVO ids and has ability for more sources and validation.

### LINKED ISSUE #120: Validate that the LLM outputed ENVO IDs resolve to a real ID
## Problem

For each slot, the suggestor hands the model a precomputed allowed-value set (`SchemaContextBuilder.format_env_triad_context`) but uses it only as an advisory prompt hint. Nothing checks the model's output: `validate_output` does Pydantic structural validation only, and the slot regex checks CURIE shape, not correctness. So the model authors CURIEs freely, with no check that a value is a real ontology term, that its label matches its CURIE, or that it fits the sample.

Result: it can emit label/CURIE pairs that disagree, and when the ideal term is not in the (interface-constrained) picklist, it fabricates one.

Pre-existing code (identical on main and the #91 branch: 0 diff in `env_triad_recommendation.py`, `schema_context.py`, `models/schema.py`, `utils/utils.py`), so this is a main issue, independent of the agentic work.

## Evidence (env_triad single-pass path)

The SPRUCE peatland fixture produced three values whose CURIEs resolve to unrelated terms:

<details><summary>OLS4 verification</summary>

| Emitted | ENVO id actually resolves to |
|---|---|
| `peatland biome [ENVO:01000253]` | freshwater river biome |
| `bog [ENVO:00000048]` | navigation canal |
| `peat [ENVO:00002000]` | slope |

```
curl -s "https://www.ebi.ac.uk/ols4/api/ontologies/envo/terms?obo_id=ENVO:01000253" | jq -r '._embedded.terms[0].label'
# -> freshwater river biome
```

The soil `env_broad_scale` picklist has no peatland biome, so the model went off-menu and invented the term and its id. `EnumValueInfo` also drops `pv.meaning`, so the code never holds the authoritative CURIE to cross-check.
</details>

## Scope: not just env_triad

The suggestor emits many fields. This should apply to every suggested slot whose schema range is an enum or a CURIE/label type. env_triad is the pilot.

## Constraints

- Do not hard-require enum membership: the value sets do not anticipate all experiment types, so membership is a preferred prior, not a gate.
- Do not bail to empty when there is no perfect match: generalize to a valid, more general ancestor term.
- Prefer semantic correctness over lexical match (the ideal value is both).

## Proposed solution

1. The model proposes a label or description per slot, not a CURIE. The code attaches the authoritative CURIE. Keep `pv.meaning` in `EnumValueInfo`.
2. Resolver reaches the whole ontology, not just the enum: exact enum match first (verified `meaning`), else OLS4 per-ontology embeddings search, then generalize to the nearest valid ancestor when a confident specific match is absent.
   ```
   curl -s "https://www.ebi.ac.uk/ols4/api/v2/ontologies/envo/classes/llm_search?q=<desc>&model=llama-embed-nemotron-8b_pca512&rows=8"
   ```
   Reference client: `nmdc-schema/src/scripts/ols4_embeddings_search.py`.
3. Validation layer: verify every emitted value is a real term with matching id/label; flag out-of-enum terms rather than silently reject; never fabricate.
4. Apply generically by slot range: enum, CURIE/label, or pattern/typed.
5. Flag expected-but-absent terms instead of fabricating. When a concept one would expect in an ontology has no term at the needed specificity, do not invent one and do not silently drop the gap: emit a needs-new-term signal (the concept, the valid ancestor used as a fallback, and a candidate-for-NTR flag), and cross-reference any known open request. This makes the suggestor a detector of ontology gaps that can feed term requests.

## Expected-but-absent terms: the peatland case

The peatland fabrication is exactly this pattern. There is no wetland/peatland/mire biome in ENVO (verified: no exact term for `peatland biome`, `mire biome`, `bog biome`, `wetland biome`), so the model invented `peatland biome [ENVO:01000253]` (which resolves to freshwater river biome). The gap is already tracked upstream as a New Term Request: EnvironmentOntology/envo#1666 (`NTR: wetland biome (biome for peat-forming wetlands)`, https://github.com/EnvironmentOntology/envo/issues/1666).

Correct behavior here: generalize `env_broad_scale` to a valid biome that does exist (for example `tundra biome [ENVO:01000180]` or `terrestrial biome [ENVO:00000446]`) and flag the sample as pending EnvironmentOntology/envo#1666, rather than mint a fake CURIE.

## Out of scope (tracked elsewhere)

- Semantic-appropriateness scoring (is a valid term the right one) belongs in nmdc-ai-eval subsumption scoring.
- Malformed example CURIE: #119. Bad skill examples: PR #91 comment. Test markers: separate issue.

sujaypatil96:
Closes #109.

The env-triad skill identified a MIxS interface, dumped that interface's value set, and let the model pick a `"label [CURIE]"` string. Nothing checked the result. And measured against the installed schema, **only 4 of the 12 env-triad-bearing extensions ship a value set at all** — soil, water, sediment, plant-associated. For air, built environment, biofilm, host-associated, both hydrocarbon extensions, misc-envs, and wastewater/sludge there was nothing to dump, so the model free-associated ENVO IDs with zero grounding. That is where the issue actually bites.

Suggestions now resolve in three tiers — curated value set → verified ENVO term outside it → broadest defensible term — and every value is validated before it is emitted.

## On the issue's security question

> We need to be conscious about just pointing agent to a URL - opens agent to security risks.

The agent is never pointed at a URL. `scripts/build_envo_index.py` fetches one hard-coded URL **at build time** and commits `data/envo_index.json.gz` — 244 KB, 2 ms to load, standard library only, byte-reproducible so an unchanged ENVO release produces no diff. Runtime reads the committed artifact and does no network I/O. Refresh deliberately with `make build-envo-index`.

`oaklib`/`runoak`, as used in nmdc-ingest-agent, was the alternative. It fits a curator's laptop; it does not fit a container under pip-audit/deptry/Dependabot — ~15 MB first-run download, 77 MB on disk, and a large transitive tree for four functions. **The practice carries over, the implementation does not.** Net dependency change: zero.

## Three findings worth flagging

**The schema declares three value patterns, not one.** 28 slots are ENVO-only; `HostAssociatedInterface`'s local scale and medium also admit UBERON; plant/soil/water admit PO. Validation reads the real pattern per slot, which makes host anatomy legitimately available — something the skill never surfaced before.

**Two dead CURIEs.** `ENVO:02000139` (desert spring) and `ENVO:02000145` (subterranean lake) sit in the water value sets and no longer exist in ENVO or OLS4. The gate rejects them. Probably worth an upstream issue on submission-schema.

**Two silent bugs in `MixsExtensions`.** The portal sends `"miscellaneous natural or artificial environment"`; the enum had `"artifical"` — so every misc-envs submission was dropped with a warning and got no value set. `WastewaterSludgeInterface` was missing entirely. Both fixed, with a test asserting the roster stays complete.

## Calibration

The anchor gate is deliberately softer than nmdc-ingest-agent's in two places, because a strict version would reject values NMDC's own schema allows:

- **Advisory for `env_local_scale`** — those value sets carry 13–16 terms per package outside `ENVO:01000813` (`aquifer`, `farm`, `fen`, `litter layer`). Hard for `env_broad_scale` and `env_medium`, whose value sets are 100% anchor-clean.
- **Never applied to a curated value** — `rhizosphere` is a curated plant medium that ENVO classes as a system. The gate polices expansion, not the schema. Existence and obsolescence still apply everywhere.

A parametrized test asserts no curated value in any of the four value sets is rejected except the two dead ones, so a schema bump that breaks that fails CI instead of silently dropping suggestions.

## Where the gate actually lives

`enforce_env_triad_values` runs on **both** paths — the programmatic pipeline and `agentic()`. It corrects label drift in place and replaces a hard failure with the extension's generic fallback, so no value reaches a caller unchecked. The skill's own validation step is guidance for getting it right the first time; this is what makes it a guarantee.

Previously it ran only on the programmatic path, and the agentic path returned whatever the model emitted. Closing that was worth more than the skill-side script I had built to encourage the same discipline, which is now removed — a step the model can skip is not a gate.

One consequence worth reviewing: `agentic()` has no interface list to pass, so unknown means *every extension is possible* rather than *the strictest one applies*. Otherwise a legitimate UBERON host medium would be rejected for want of context. Existence, obsolescence, label and anchor checks are unchanged; only the prefix check goes lenient.

That change surfaced a bug worth naming — interfaces iterate in enum order, `BuiltEnvInterface` is first and has no value set, so taking the first merely-valid result labelled every curated value as `envo_expansion`. Selection now prefers an interface that found the value in its curated set, with a test pinning it.

`SKILL.md` states the tiers and calls the index directly; detail lives in `refs/` per progressive disclosure, which matches the four existing skills in this repo.

## Room for domain experts

The environment-specific reference is split one file per MIxS extension under `refs/extensions/`, so a soil specialist and a water specialist never collide in one diff, and the agent loads only the environment its samples came from — 4.5 KB for the roster plus one extension against 12 KB for all twelve, a gap that widens with every contribution. Filenames derive from the interface class, so the path follows from the extension the skill already identified.

Each file carries a fixed skeleton with **Domain notes** and **Worked examples** marked undocumented — a visible target rather than a blank page. `CONTRIBUTING.md` has a guide aimed at people who do not write Python.

The point of leverage is `tests/test_env_triad_references.py`: every ENVO term written into these files must exist in the pinned release, not be obsolete, carry ENVO's official label, and be the right kind of term for any slot its line names. A reviewer does not have to know ENVO by heart. I checked it against all three failure modes — and the trial caught me out, which is the argument for it: the CURIE I invented as obviously-fake, `ENVO:00005777`, is real. It is *steppe soil*, not the *paddy soil* I had labelled it, and the label check is what flagged it.

## Verification

- `make lint`, `uv run ruff format --check`, `make check-deps`, `uv run pytest` — all clean, no new dependency.
- Wheel ships the artifact (`data/envo_index.json.gz` verified present).
- **Not verified: no live LLM run.** No GCP/PNNL credentials available, so the four `-m integration` env-triad tests skip. I exercised the full pipeline with a stubbed `ConversationManager` instead — a wastewater study (previously unroutable) yields three grounded ENVO values, label drift auto-corrected, bad CURIE replaced by the seed fallback. **A real integration run is worth doing before merge.**

## Left alone

`get_env_triad_recommendation(interface_names=None)` still defaults to all interfaces (~51 KB of value sets). Narrowing it would help but changes behavior for existing ETL callers, so it is documented in the docstring rather than changed unasked. The skill instructs against the fallback.

**No `CODEOWNERS`.** Per-file ownership under `refs/extensions/` is the natural fit once the domain experts are confirmed — a soil PR requesting the soil reviewer and nobody else. It belongs in the PR that names them, not here as a fully commented-out file that assigns no one.

When a value fails on the agentic path the fallback is the bare anchor (`environmental material [ENVO:00010483]`) rather than the extension's sensible default, because nothing tells `agentic()` which extension is in play. Threading `interface_names` through would fix it; it is not needed for correctness.


src/nmdc_metadata_suggestor_ai_tool/envo_index.py
```diff
@@ -0,0 +1,627 @@
...
+
+Usage mirrors :class:`~nmdc_metadata_suggestor_ai_tool.schema_context.SchemaContextBuilder`::
+
+    from nmdc_metadata_suggestor_ai_tool.envo_index import get_envo_index
+
+    index = get_envo_index()
+    index.search("rhizosphere", slot="env_local_scale")
+    index.validate("soil [ENVO:00001998]", "env_medium")
+"""
+
```
sujaypatil96: We are not adding oaklib as a dependency in this repo b/c we want to avoid having the agent built as a containerized service in this repo pull in relatively large artifacts at runtime. There are four operations that we need oaklib (runoak commands) for - look up a term, check it isn't obsolete, confirm its label, and test whether it descends from an anchor class. Those four operations have been implemented in this module and they work against a pinned release of ENVO - you can find that at this path: `src/nmdc_metadata_suggestor_ai_tool/data/envo_index.json.gz`

.claude/skills/env-triad/refs/extensions.md
```diff
@@ -0,0 +1,48 @@
...
+| `soil` | `SoilInterface` | 52/83/85 | [extensions/soil.md](extensions/soil.md) |
+| `water` | `WaterInterface` | 56/86/96 | [extensions/water.md](extensions/water.md) |
+| `air` | `AirInterface` | none | [extensions/air.md](extensions/air.md) |
+| `built environment` | `BuiltEnvInterface` | none | [extensions/built-env.md](extensions/built-env.md) |
+| `host-associated` | `HostAssociatedInterface` | none | [extensions/host-associated.md](extensions/host-associated.md) |
+| `hydrocarbon resources-cores` | `HcrCoresInterface` | none | [extensions/hcr-cores.md](extensions/hcr-cores.md) |
+| `hydrocarbon resources-fluids_swabs` | `HcrFluidsSwabsInterface` | none | [extensions/hcr-fluids-swabs.md](extensions/hcr-fluids-swabs.md) |
+| `microbial mat_biofilm` | `BiofilmInterface` | none | [extensions/biofilm.md](extensions/biofilm.md) |
+| `miscellaneous natural or artificial environment` | `MiscEnvsInterface` | none | [extensions/misc-envs.md](extensions/misc-envs.md) |
+| `wastewater_sludge` | `WastewaterSludgeInterface` | none | [extensions/wastewater-sludge.md](extensions/wastewater-sludge.md) |
```
sujaypatil96, line 26: Guidance for each of the MIXS extensions (other jargon used - "domains") is split by environment, ready for experts. We are adding one file per sample type , each with a section for domain knowledge and worked examples. For example a soil biologist can edit and record specialized knowledge in `soil.md`.

scripts/build_envo_index.py
```diff
@@ -0,0 +1,213 @@
...
+Then review the diff in ``envo_version`` and the term counts before committing.
+
+Why an ENVO-only DAG: the ENVO release graph mixes in imported classes from
+other ontologies (FOODON, CHEBI, PATO). The env triad slots accept only
+``ENVO:`` CURIEs, so those imports are dropped from the index. To keep the
+subclass hierarchy connected after dropping them, each ENVO class stores its
+*nearest ENVO ancestors* -- walking up through any non-ENVO intermediates --
+rather than its literal parents. Anchor membership is computed on the full
+graph before the trim, so the two agree exactly.
+"""
```
sujaypatil96: We use code in this module to fetch ontology data over the network. It only runs at build time, never at request time. The agent is never handed a URL, it only reads the committed artifact. Every time there's a new release of ENVO we can build and pin a new ENVO release artifact in this repo by running this command: `make build-envo-index`

src/nmdc_metadata_suggestor_ai_tool/envo.py
```diff
@@ -0,0 +1,627 @@
...
+            "ENVO:00002001",  # waste water
+            "ENVO:00002018",  # sewage
+            "ENVO:00002059",  # biosolids
+        ),
+    },
+    # HostAssociatedInterface and MiscEnvsInterface intentionally have no seeds:
+    # their sampled material is not constrained to any one ENVO subtree, so
+    # expansion runs over the whole environmental-material anchor.
+}
+
```
sujaypatil96, line 109: ENVO subtrees that we look under for extensions that have no curated value set (we have value sets only for air, water, soil, sediment). The anchor classes were chosen by LLM judgement, so this would be a good section for a domain expert to eyeball.

CC: @mslarae13

src/nmdc_metadata_suggestor_ai_tool/models/llm_output.py
```diff
@@ -20,6 +20,15 @@ class MetadataFieldSuggestion(BaseModel):
...
     ) = Field(default="", description="The recommended value for the metadata field")
+    source: str | None = Field(
+        default=None,
+        description=(
+            "Where the value came from, for env triad suggestions: "
+            "'submission_enum' (in the schema's curated value set for this MIxS "
+            "extension), 'envo_expansion' (a verified ENVO term outside that set), "
+            "or 'generalized' (broadest defensible term). Absent for other fields."
+        ),
+    )
```
sujaypatil96, line 89: We are using this optional field to record provenance -- about which tier (documented in SKILL.md) a suggested value came from.

- **_submission_enum_** if it's in the schema's curated list for that extension
- **_envo_expansion_** if it's a verified ENVO term outside that list
- **_generalized_** if we fell back to the broadest defensible term

src/nmdc_metadata_suggestor_ai_tool/llm_client.py
```diff
@@ -407,7 +410,9 @@ async def agentic(
                     if isinstance(event, SystemMessage) and event.subtype == "init":
                         session_id = event.data["session_id"]
                     elif isinstance(event, ResultMessage):
-                        result = self.unwrap_structured_output(event.structured_output)
+                        result = enforce_env_triad_values(
+                            self.unwrap_structured_output(event.structured_output), None
+                        )
```
sujaypatil96: By adding this we are enabling the agentic pipeline to check its own work. Previously, this path returned whatever the model produced (unchecked) - only the non-agentic pipeline validated anything. Running `enforce_env_triad_values` here means every suggestion is checked in code regardless.

src/nmdc_metadata_suggestor_ai_tool/constants.py
```diff
@@ -134,3 +134,33 @@
...
+    "env_broad_scale": "ENVO:00000428",  # biome
+    "env_local_scale": "ENVO:01000813",  # astronomical body part
+    "env_medium": "ENVO:00010483",  # environmental material
+}
+
+# Anchor membership is a hard gate only where the submission schema's own value
+# sets are anchor-clean. Every curated env_broad_scale and env_medium value sits
+# under its anchor; env_local_scale value sets carry 13-16 terms per package that
+# do not (aquifer, farm, fen, litter layer), so there it is advisory only.
+HARD_ANCHOR_SLOTS: frozenset[str] = frozenset({"env_broad_scale", "env_medium"})
```
sujaypatil96, line 204: The anchor check is a hard failure for `env_broad_scale` and `env_medium`, but only a warning for `env_local_scale`.

.claude/skills/env-triad/refs/extensions.md
```diff
@@ -0,0 +1,48 @@
+# MIxS extension reference
+
+Twelve MIxS extensions carry env triad slots. Only four ship a curated value set, so for the other eight ENVO expansion is not a fallback — it is the only grounding available.
+
+Portal names come from `HARMONIZER_TEMPLATES` in nmdc-server; `MixsExtensions.map_to_interface_name` accepts either of the first two columns.
+
+## Roster
+
+**Read only the file for the extension in play.** Each holds that extension's evidence signals and domain knowledge; loading all twelve wastes context on environments the samples are not from.
```
caufieldjh: Is it possible or likely for multiple extensions to be in use? I would think yes.

.claude/skills/env-triad/SKILL.md
```diff
@@ -31,84 +31,98 @@ For sample level suggestions, the suggestion must be tied to an ID. Copy the ID
...
-- Every `reason` must cite specific input text (exact short quote up to 12 words, or paraphrase with source label such as "abstract", "study description", "sample field")
+- Every `reason` must cite specific input text (exact short quote up to 12 words, or paraphrase with a source label such as "abstract", "study description", "sample field")
 - No generic domain justifications without citing input text
 - Do not reference schema package names (e.g. `SoilInterface`) in reasons
+
+---
+
+## Step 6 — Validate before emitting
+
+Check every value before emitting it. A value that fails does not get emitted — drop a tier and try again.
```
caufieldjh, line 129: when should the process of "drop a tier and try again" stop?

scripts/build_envo_index.py
caufieldjh, line 1: Some of the functions in this file are so general-purpose (like iri_to_curie() or load_graph()) that it seems strange to have them live here. Move into src/utils/utils.py?

CONTRIBUTING.md
caufieldjh, line 1: Thanks for updating this @sujaypatil96

.claude/skills/env-triad/refs/extensions.md
hesspnnl: yes very possible

sujaypatil96: Thanks @caufieldjh! yes, agreed, the wording is confusing/misleading here. I'll fix that. The code however does already handle a list of extensions throughout so we're good on that front - don't need any code changes.

.claude/skills/env-triad/SKILL.md
sujaypatil96: In the code, we make a call `generic_fallback()` (see https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/blob/109-envo-term-expansion/src/nmdc_metadata_suggestor_ai_tool/envo_index.py#L599) which guarantees that a value is returned, and so technically the process of "drop a tier and try again" stops after tier 3 (by construction). The notes in the skill files don't explicitly say that this is where it terminates, but a good clarifying statement to add to help clarify ambiguity for a reader.

scripts/build_envo_index.py
sujaypatil96: Agreed about `iri_to_curie()` -- that can definitely be moved to `src/utils/utils.py`, but `load_graph()` on the other hand is OBO graph-specific. Also these methods are currently only used for building the pinned ENVO artifact and so it makes sense for them to live in `scripts/` b/c that folder is not included in the wheel (whereas everything under `src/` is).

sujaypatil96: Addressed in [112db5a](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/commits/112db5af1f87ed589e0bea84302e84f81a14beff)

.claude/skills/env-triad/refs/extensions.md
sujaypatil96: Addressed in [381816a](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/commits/381816a32213647b0217a08875b7a3e592311170)

.claude/skills/env-triad/SKILL.md
sujaypatil96: Addressed in [a85ac16](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/commits/a85ac16904d9e6eef0c0427a8d96beeec6a0ea1a)

hesspnnl (changes requested):
As discussed in the squad meeting 8/18 - Alicia and I are wondering about the choice to keep a local copy of oak data and functions that we need. i can foresee this adding another maintenance step and potentially causing issues with mismatched versions of the data/functions. Alicia also mentioned potentially using the NMDC envo db as well. Requesting changes to this part of the MR. Thanks!

turbomam (commented):
Reviewed against https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/122, which is the issue I filed about valid-but-wrong-concept suggestions, and focusing on the examples. Everything below was run against `45a46e5`.

**The examples are clean.** I pulled all 9 unique `label [CURIE]` pairs out of `.claude/skills/env-triad/` and resolved each against OLS4. All 9 labels match their CURIE and none is obsolete. Worth saying explicitly because this is where the previous skill work went wrong: on https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/91 four of six examples had IDs that resolved to unrelated terms.

**The anchor gate does real work.** I ran the original SPRUCE failure through it: `peatland biome [ENVO:01000253]`, whose CURIE actually resolves to *freshwater river biome*, is rejected for `env_medium` because it is not under `environmental material`. That is the reference-integrity problem solved, and closing https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/120 against this PR is right.

**It does not close 122, and I do not think it was meant to.** Anchors constrain the category, not the choice within it. Both of the failures I filed 122 for still pass:

| case | under the slot's anchor | gate |
|---|---|---|
| a wrong-but-valid biome in `env_broad_scale` | yes | passes |
| `surface water [ENVO:00002042]` as `env_medium`, duplicating local scale | yes | passes |
| SPRUCE label/CURIE mismatch | no | rejected |

Cross-slot duplication is not checked at all. So 122 should stay open after this merges. Flagging it only so the linkage is not assumed.

**One thing I would like your read on, and it is the only substantive question I have.**

Both agentic call sites pass `interface_names=None` unconditionally:

```python
result = enforce_env_triad_values(
    self.unwrap_structured_output(event.structured_output), None
)
```

while `env_triad_recommendation.py:172` passes the real names. The docstring says `None` is deliberate, so an unknown extension does not get the strictest interface. That is reasonable when the extension is genuinely unknown. On the agentic path it is passed even when the submission tells us what the extension is.

It costs more than permissiveness, because the `env_medium` value sets are completely disjoint:

```
SoilInterface      85 values, 85 unique to it
WaterInterface     96 values, 96 unique to it
SedimentInterface  11 values, 11 unique to it
PlantAssociated    28 values, 28 unique to it
```

Zero overlap. So a water-only value suggested on a soil submission:

```
'acidic water [ENVO:01000358]'
  non-agentic (interface_names=['SoilInterface']) -> source = 'envo_expansion'
  agentic     (interface_names=None)              -> source = 'submission_enum'
```

Same value, different provenance. The agentic path reports it as coming from the submission schema's curated set when it came from a different extension's set. Given you are using `source` to track provenance, that seems worth threading the interface through rather than passing `None`. Possibly I am missing a reason it is not available at that point.

**Smaller: the gate is wired in but nothing holds it there.** I removed the call at `env_triad_recommendation.py:172` and ran the suite: 320 passed. Same for both agentic call sites. The tests exercise `enforce_env_triad_values` directly, so all three call sites can be deleted silently. I hit exactly this in my own PR today and it is easy to miss, so no criticism intended, but a test asserting the gate is reached would be cheap.

**Process note:** this still shows `CHANGES_REQUESTED` from the 2026-08-19 review about keeping a local copy of oak data, and `45a46e5` ("resolve ENVO through oaklib instead of a pinned index") landed after it. If that addressed the objection, the review may just need clearing. Not mine to dismiss.

Not approving or requesting changes, since the open question above is Sujay's call and the stale review is Olivia's.

src/nmdc_metadata_suggestor_ai_tool/models/llm_output.py
hesspnnl: how would this expand or be available for other slot value suggestions? env triad is not the only slot on our mind for value suggestions

README.md
```diff
@@ -35,6 +35,24 @@ Environment variables are loaded from a `.env` file in the project root via
 set in your shell take precedence over `.env` values (`override=False` is the
 default).
 
+### Agent permissions
+
+`ConversationManager.agentic()` reads `.claude/settings.json` for its tool allowlist. Without it
```
hesspnnl, line 40: ahh i like this solution. i tried to address the same in #139 but this solution seems more elegant.

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
```diff
@@ -56,8 +56,14 @@
 - env_local_scale: This field should capture the more specific local environment of the sample (e.g., "agricultural field [ENVO:00000114]", "rhizosphere [ENVO:00005801]", "aquifer [ENVO:00012408]").
 - env_medium: This field should capture the medium in which the sample was collected ("soil [ENVO:00001998]", "acidic water [ENVO:01000358]", "alluvial paddy field soil [ENVO:00005759]").
 
-The values for these fields should be chosen from the enumerations in the NMDC schema and should be based on the content of the information provided.
-The format of the value should be in the form of "label [CURIE]", where "label" is the human-readable term and "CURIE" is the corresponding identifier from the NMDC schema.
+The format of the value must be "label [ENVO:NNNNNNN]", where "label" is the term's official ENVO label. Values with any other prefix are rejected by the submission portal.
```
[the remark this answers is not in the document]
hesspnnl: @sujaypatil96 is this still relevant?

src/nmdc_metadata_suggestor_ai_tool/llm_client.py
```diff
@@ -344,6 +359,108 @@ def unwrap_structured_output(raw: Any) -> LLMOutput:
                         continue
         raise ValueError(f"Could not extract LLMOutput from structured_output: {raw}")
 
+    @staticmethod
+    def structured_output_from_tool_use(event: AssistantMessage) -> dict[str, Any] | None:
+        """The payload the agent passed to the StructuredOutput tool, if it called it.
+
+        ``ResultMessage.structured_output`` has come back empty on runs where the agent did
```
hesspnnl: this docstring does not do a good job conveying the purpose "where the agent did call the tool with a complete answer" - does this mean agent called a tool and received an answer and this interferes with the structured output resp?

src/nmdc_metadata_suggestor_ai_tool/models/llm_output.py
hesspnnl: this comment still not addressed

hesspnnl (changes requested):
still some outstanding issues. sorry for being picky

turbomam (commented):
Claude's report from my analyses:

`0763d8d` fixes a problem from my earlier review (https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144#pullrequestreview-5010655951): the model suggesting the same term, `surface water [ENVO:00002042]`, for both `env_local_scale` and `env_medium` on one sample. Now one of the two gets replaced.

The agent outputs you attached in https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144#issuecomment-5431919178 (agent_trace_biosample.json and agent_trace_15_bioscales_biosamples.json) also look good. `tundra biome` / `glacier` / `cryosol` for an Arctic glacier soil sample is right, and every suggested term exists in ENVO, is not obsolete, and has the correct label.

Six suggestions.

**1. Check `env_local_scale` values against an expression, not a single class.** A term should be a descendant of astronomical body part (ENVO:01000813), but not a descendant of biome (ENVO:00000428) and not a descendant of environmental material (ENVO:00010483).

The two exclusions are the point. Astronomical body part has 1,731 descendants in the ENVO release the tool uses (2025-10-20), and they include all 128 biomes and 151 environmental materials. That is why `soil [ENVO:00001998]` and `tundra biome [ENVO:01000180]` currently pass as `env_local_scale` values with no warning, even though one is a material and one is a biome. Adding the exclusions breaks nothing: I checked every curated `env_local_scale` value in the soil, water, sediment and plant-associated enums of nmdc-submission-schema, 284 in all, and none is a biome or an environmental material.

The expression matters most for the eight sample types whose enums have no `env_local_scale` values at all: built environment, air, host-associated, biofilm, both hydrocarbon resources, misc envs and wastewater sludge. For those, the model picks from all of ENVO, and this check is the only thing filtering what it picks.

There are 25 curated values the expression would wrongly flag (`marsh`, `peatland`, `fen`, `glacier`, `aquifer`, `vadose zone` and others). Those reflect missing links in ENVO itself. For example `fen [ENVO:00000232]` is classified under wetland ecosystem, which ENVO does not place under astronomical body part, even though it places terrestrial ecosystem there. I can file that with ENVO. Until such fixes land, keeping this check advisory for `env_local_scale`, as you already do, is right.

**2. Add two more allowed ancestor classes: vegetation layer (ENVO:01000355) and environmental zone (ENVO:01000408).**

nmdc-submission-schema's definition of `env_local_scale` lists four example values: `canopy`, `herb and fern layer`, `understory` and `shrub layer`. Your check warns on three of them, because they descend from vegetation layer, not from astronomical body part. So the schema's own examples fail the tool's check. (The fourth, `canopy [ENVO:00000047]`, fails for a better reason: that identifier is obsolete in current ENVO, and the schema and ENVO's wiki both still publish it. Your validator catching it is correct. I can report it upstream.)

The same slot definition links ENVO's guidance page, https://github.com/EnvironmentOntology/envo/wiki/Using-ENVO-with-MIxS. Its worked example for a forest sample uses those same vegetation layer terms for `env_local_scale`, and it states the idea behind item 1 in plain terms: `env_local_scale` should name countable things (a rock, a cave, a hydrothermal vent) and `env_medium` should name masses (soil, water).

Environmental zone would add 19 terms that astronomical body part does not cover, such as `atmospheric zone`, `populated place` and `ice accumulation zone`. No curated enum contains them, but they are sensible `env_local_scale` suggestions for the eight sample types that have no enum.

**3. Give empty values the same treatment as invalid ones.** `enforce_env_triad_values` skips a suggestion whose value is `""` (`envo.py:704`), so its own docstring's promise, "callers never receive an invalid value and never receive an empty one", does not hold:

```
enforce_env_triad_values(one_suggestion(""), ["SoilInterface"])
-> value='' source=None
```

Codex flagged the same line: https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144#discussion_r3780663057

**4. Let `source="generalized"` survive validation.** When the model deliberately answers with a broad term and labels it `generalized`, validation overwrites that label, because `ValidationResult.source` can only return `submission_enum` or `envo_expansion`:

```
in : air [ENVO:00002005]  source='generalized'
out: air [ENVO:00002005]  source='envo_expansion'
```

The result is that `generalized` only ever appears on values the validator itself substituted, never on ones the model chose. Codex: https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144#discussion_r3780663071

**5. Pass the real interface names at `llm_client.py:462` instead of `None`.** With `None`, a value counts as `submission_enum` if it is in any sample type's enum, not the submission's own. To show the effect, I ran the suggestions from your attached agent_trace_15_bioscales_biosamples.json back through `enforce_env_triad_values` with three different `interface_names` arguments:

| value | `None` (what your run used) | `["SoilInterface"]` | `["PlantAssociatedInterface"]` |
|---|---|---|---|
| `plantation [ENVO:00000117]` | submission_enum | envo_expansion | submission_enum |
| `root [PO:0009005]` | submission_enum | envo_expansion | submission_enum |
| `bulk soil [ENVO:00005802]` | submission_enum | submission_enum | envo_expansion |

Same values, different provenance labels, depending only on an argument that is currently always `None` on the agentic path.

**6. Re-generate the attached outputs with validation on.** In agent_trace_15_bioscales_biosamples.json, `plantation [ENVO:00000117]` appears five times, three labelled `submission_enum` and twice `envo_expansion`. `enforce_env_triad_values` always gives the same label to the same slot and value, so it cannot have produced that file; the attachment looks like raw model output from before validation. Running the file through validation also replaces `rhizosphere [ENVO:00005801]`, which is not an environmental material.

Two smaller things. `system_prompt.py:59` tells the model the portal rejects every prefix except ENVO, but 26 of the 28 curated plant-associated `env_medium` values are Plant Ontology terms (https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144#discussion_r3780663064; @hesspnnl asked about the same line today at https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144#discussion_r3905667165). And the PR currently has merge conflicts, which block CI.

All measurements at `22a6c65`.

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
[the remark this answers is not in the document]
hesspnnl: we need to modify the guidance to allow schema approved PO and UBERON values

src/nmdc_metadata_suggestor_ai_tool/envo.py
```diff
@@ -0,0 +1,627 @@
...
+        list(interface_names) if interface_names else list(MixsExtensions.__members__)
+    )
+    fallback_interface = interface_names[0] if interface_names else ""
+
+    for suggestion in output.metadata_fields:
+        slot = suggestion.field_name
+        if slot not in ENV_TRIAD_SLOTS or not isinstance(suggestion.value, str):
+            continue
+        if not suggestion.value:
+            continue
```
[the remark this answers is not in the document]
hesspnnl: there are three tiers in which the agent will look for values. it retains provenance for each of the sources and falls back to general knowledge.

src/nmdc_metadata_suggestor_ai_tool/models/llm_output.py
hesspnnl: harry recommended using hooks - https://www.mindstudio.ai/blog/claude-code-skills-vs-hooks-difference. the context injection in this case is if the suggestion is env triad, the source needs to come from these three options.

README.md
sujaypatil96: Yup, I left a comment about the same on #139. See https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/139#discussion_r3906946934

src/nmdc_metadata_suggestor_ai_tool/llm_client.py
sujaypatil96: Addressed in [b5c6df0](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/commits/b5c6df02c2f9b5637558ec0d3f1edc2d41cf02c5) -- added example snippet of better wrapping for `StructuredOutput` with `metadata_fields` payload inside it.

sujaypatil96:
Addressing the comments from @turbomam's review [here](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144#pullrequestreview-5081936188):

- [x] Fixed `enforce_env_triad_values()` so it checks for empty values too. Non-string values were being skipped too, so added that as well (addressed in [2269774](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/commits/2269774968e96aeb228dfd1046cf6c239a8c5204)) -- point 3
- [x] Checking accepted env triad values against each interface's fallback (addressed in [a5379f7](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/commits/a5379f790195f9cd8e0fd15f0cd46d47db2714c1)) -- point 4
- [x] `agentic()` and `finalize_result()` now take interface_names (addressed in [05f33b6](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/commits/05f33b60b7369946b2bbc7855eb607211d8f8314)) -- point 5

sujaypatil96:
Added some commits above to resolve any tool calling errors noticed in agent traces.

sujaypatil96:
This PR has addressed all review comments from @hesspnnl and @turbomam that were discussed at the squad sync call on 9/1.

sujaypatil96:
JSON output from agent traces -
* [agent_trace_biosample.json](https://github.com/user-attachments/files/31722682/agent_trace_biosample.json)
* [agent_trace_15_bioscales_biosamples.json](https://github.com/user-attachments/files/31722681/agent_trace_15_bioscales_biosamples.json)

scripts/build_envo_index.py
hesspnnl: outdated - no longer doing this

src/nmdc_metadata_suggestor_ai_tool/system_prompt.py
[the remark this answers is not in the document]
sujaypatil96: Addressed in [0f7f860](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/commits/0f7f8602eeb3806af442d3e869592c2487a547ba)

src/nmdc_metadata_suggestor_ai_tool/env_triad_recommendation.py
```diff
@@ -16,6 +20,32 @@
...
+        "# Verified ENVO terms (ENVO expansion tier)",
+        "Use these only when the schema enumerations above have nothing appropriate, "
+        "or when no enumeration was supplied for this MIxS extension.",
+        "",
+        f"## env_broad_scale — every ENVO biome ({len(biomes)}; this is the complete set)",
+        ", ".join(biomes),
+    ]
+    for name in interface_names or []:
+        for slot in ("env_local_scale", "env_medium"):
+            if index.expansion_seeds(name, slot):
```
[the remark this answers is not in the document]
sujaypatil96: Addressed in [f19a671](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/144/commits/f19a671e97f7e73618c10e55dcaa534ce2f0cd2b)

hesspnnl (approved):

PR #145, opened 2026-08-11
Rank supplements by file kind before applying the caps

### LINKED ISSUE #142: Supplement file limit keeps whatever comes first in the archive, not the most useful files
The supplement retriever keeps at most 10 files per article. It reads files in the order they appear in the downloaded archive and stops once it has 10. It does not prefer one file type over another.

The skill document at `.claude/skills/supplement-retrieval/SKILL.md` says spreadsheets are the highest-value type, because they usually hold per-sample metadata, and that documents such as PDFs come second.

Those two things disagree. Here is what happens on https://doi.org/10.1038/s41564-020-00861-0, which is the example DOI in the skill document.

The archive holds 52 files. Nineteen of them are spreadsheets or PDFs. The retriever kept 10 and dropped the other 9. Of the 10 it kept, 9 are spreadsheets and 1 is a PDF. Of the 8 dropped for hitting the limit, 7 are spreadsheets and 1 is a PDF. One further PDF was dropped for exceeding the single-file size limit.

So a PDF occupied one of the 10 slots while 7 spreadsheets were discarded, purely because of where each file sat in the archive.

Suggested change: sort the candidate files by type before applying the limit, so spreadsheets fill the slots first and documents take whatever remains.

The code is `select_members` in `src/nmdc_metadata_suggestor_ai_tool/publication_ingestion/supplements/shared.py`. It is a single loop over the archive members with no sorting, and the limit check sits inside that loop.

Found while reviewing https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/132 (Skills and functions for supplement retrieval). It is not a defect in that pull request and does not block it.

caufieldjh:
Fixes #142 

Supplementarty files retrieved using the scripts added by #132 competed for max_files/max_total_bytes by listing position, so a record whose repository listed figures and PDFs first could spend the whole budget before reaching a single spreadsheet, though these are the files that may actually carry per-sample metadata.

Add `KIND_PRIORITY` to file_kinds: a total order over every SupplementKind, data-like first (tabular > document > archive > other > sequence > image > media), with kind_rank() as its sort key. A test asserts the tuple
covers the whole enum, so a new kind must be placed deliberately rather than sorting last by accident.

Apply it in the two places a cap is enforced:

* select_members stable-sorts the listing before spending the budget, so every source inherits the ranking. Member.read stays lazy, so this reorders names and sizes without triggering a download.
* merge_results ranks the cross-source union the same way, letting a linked dataset's spreadsheet outrank a hosted figure.

Both sorts are stable, so within one kind the previous order survives (source listing order, and hosted supplements ahead of linked datasets).

Priority ranks what has been fetched; it cannot reach across the fetch order, since a source that exhausts max_files leaves later sources' files undownloaded. Documented on retrieve_supplements and in the skill reference rather than fixed, as over-fetching to discard would defeat the purpose of doing a rapid check for available supplementary context.

### DIFF
--- a/.claude/skills/supplement-retrieval/SKILL.md
+++ b/.claude/skills/supplement-retrieval/SKILL.md
@@ -22,6 +22,11 @@ samples and studies. Keep, in priority order:
 Skip (low value for metadata suggestion): figures/images, audio/video, raw
 sequence data (`.fasta`, `.fastq`, `.bam`, …), and nested archives.
 
+This is a ranking, not just a filter: when a record offers more files than the
+caps allow (default `max_files=10`), the budget goes to the tabular files first
+and only the leftovers go to documents — a record with 12 spreadsheets and 3 PDFs
+returns 10 spreadsheets, not whichever 10 the repository happened to list first.
+
 ## Performance first
 
 Supplements are a *nice-to-have*, not a blocker. Make **one** attempt per DOI. If
--- a/.claude/skills/supplement-retrieval/refs/programmatic.md
+++ b/.claude/skills/supplement-retrieval/refs/programmatic.md
@@ -27,8 +27,14 @@ from nmdc_metadata_suggestor_ai_tool.file_kinds import (
     DEFAULT_USEFUL_KINDS,  # {TABULAR, DOCUMENT}
     EXTENSION_KINDS,  # extension -> SupplementKind, the full mapping
     TEXT_LIKE_EXTENSIONS,  # extensions inlined as text rather than saved
+    KIND_PRIORITY,  # kinds in the order they win a contested budget
+    kind_rank,  # SupplementKind -> sort rank (lower = more valuable)
 )
 ```
+
+`KIND_PRIORITY` orders every kind most-valuable-first — `tabular`, `document`,
+`archive`, `other`, `sequence`, `image`, `media` — and is what decides *which*
+files survive a cap, not just which kinds are eligible.
 
 `DEFAULT_USEFUL_KINDS` and `classify_file` (aliased there as
 `classify_supplement`) are also re-exported from the supplements package, so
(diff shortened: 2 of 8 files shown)


hesspnnl:
I think this seems fine. @turbomam any thoughts?

turbomam (approved):
Approving. Reviewed against `4c6018d`, which fixes https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/142, the issue I filed. It does what I was asking for and the tests hold it in place.

Sorry this sat 13 days. I was the requested reviewer and did not notice.

**I checked the claims rather than reading them.** Both cap sites really are ranked, and each is pinned by a test that fails without it. Neutered the sort key at each site in turn and ran the suite:

| mutation | result |
|---|---|
| `select_members` sort removed (`shared.py:265`) | 3 tests fail, incl. `test_europepmc_max_files_goes_to_tabular_over_documents` |
| `merge_results` sort key made constant (`retrieve.py:219`) | 1 test fails, `test_merge_results_prefers_data_like_files_across_sources` |
| neither | 263 pass |

That matters more than usual here. A ranking is the kind of thing that keeps passing after someone removes it, because the output is still a plausible list of files in some order. These do not.

**`test_kind_priority_covers_every_kind` is the best thing in the PR.** `set(KIND_PRIORITY) == set(SupplementKind)` plus the round-trip through `sorted(..., key=kind_rank)` means a new `SupplementKind` cannot quietly sort last. `kind_rank` falls back to `len(KIND_PRIORITY)` for an unknown kind, which is the safe default, and this test is what stops that default from becoming the silent normal case. I spent today finding several checks in this repo that could be deleted without any test noticing, so it is worth naming one that is built the other way round.

**On the ordering itself**, `ARCHIVE` above `OTHER` and `SEQUENCE` looks right to me for the metadata-extraction use case: a zip plausibly bundles the tables, where a bam almost never helps characterize samples. `OTHER` above `SEQUENCE` is the only one I paused on, but the comment gives the reason (delimited data under an odd suffix), and it is a judgment call rather than a defect.

Stability is claimed and it holds: both sorts are `sorted(...)` with a rank-only key, so within a kind the listing order survives, which is what keeps Table S1 ahead of Table S2.

`ruff`, `mypy src` and 263 tests clean on my checkout.

**One thing to know, not a change request.** While reviewing something else I found that `parse_supplement_captions` drops captions when the filename and the caption sit on different elements, which is common JATS: the caption is on `<supplementary-material>` and the `xlink:href` is on a child `<media>`. Filed as https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/159. It does not affect this PR, since ranking works off filenames rather than captions, but it means caption-derived signal is currently missing for some documents you might otherwise expect to rank well.

caufieldjh:
Thanks @turbomam!

PR #150, opened 2026-08-20
Fix the setup and Quickstart docs

### LINKED ISSUE #103: Make the Quickstart runnable and document what the output means
The README "Use the package in Python" example uses an empty placeholder:

```python
submission_object = {
    # NMDC submission JSON payload
}
```

so following the Quickstart runs nothing. A new user has no ready input and no runnable command.

Proposal:
- Give a copy-pasteable example that runs against real input, for example `tests/fixtures/full_submission.json`, or add a small `scripts/` runner or console command.
- Document what the output means. Per the system prompt's VALUE FIELD RULES in `src/nmdc_metadata_suggestor_ai_tool/system_prompt.py`, this task identifies which NMDC metadata fields a submitter should fill in (per biosample, via the `id` rules), each with an evidence-based `reason`. `value` is populated only when the input contains the literal value text; inferred values go in the `reason` and `value` is left empty. So empty `value` fields are expected behavior, not a bug. The README should say this plainly so users understand the output is field guidance, not value completion.

Separate observation: `test_vertexai_integration.py` asserts at least one field has a non-empty `value`, which is inconsistent with the VALUE FIELD RULES above (values are usually empty by design). That assertion looks fragile and may warrant its own look.

sujaypatil96:
Documentation fixes from getting set up with this repo for development.

- Document `CLAUDE_CODE_USE_VERTEX`, which the agentic path needs to route to Vertex. The other two Vertex variables I'd added are optional after all — the CLI defaults the region to `us-east5` and reads the project from the service account credentials — so they're out.
- Make the Quickstart runnable: load a fixture that ships with the repo instead of an empty placeholder.
- Explain the output: an empty `value` is expected, since the pipeline suggests which fields to fill in and why, not the values themselves.
- Fix the `ConversationManager` example, which omitted the required `system_prompt` argument.

Closes #103

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,11 +30,9 @@ VERTEX_PROJECT_ID=your_gcp_project_id
 GCP_REGION=us-east5
 
 # Claude Agent SDK configuration, used only by ConversationManager.agentic().
-# That path runs a Claude model through the Claude Code CLI, which reads these to
+# That path runs a Claude model through the Claude Code CLI, which reads this to
 # route to Vertex. Unused by the Gemini path.
 CLAUDE_CODE_USE_VERTEX=1
-ANTHROPIC_VERTEX_PROJECT_ID=your_gcp_project_id
-CLOUD_ML_REGION=us-east5
 
 # DOI / Publication APIs
 # Contact email used in User-Agent headers for Crossref polite pool and OpenAlex.
--- a/.env.example
+++ b/.env.example
@@ -29,6 +29,13 @@ VERTEX_PROJECT_ID=your_gcp_project_id
 # Default Vertex region (the code default is us-east5 if unset)
 GCP_REGION=us-east5
 
+# Claude Agent SDK configuration, used only by ConversationManager.agentic().
+# That path runs a Claude model through the Claude Code CLI, which reads these to
+# route to Vertex. Unused by the Gemini path.
+CLAUDE_CODE_USE_VERTEX=1
+ANTHROPIC_VERTEX_PROJECT_ID=your_gcp_project_id
+CLOUD_ML_REGION=us-east5
+
 # DOI / Publication APIs
 # Contact email used in User-Agent headers for Crossref polite pool and OpenAlex.
 # Crossref: https://github.com/CrossRef/rest-api-doc#good-manners--more-reliable-service
--- a/README.md
+++ b/README.md
@@ -67,25 +67,35 @@ default).
    ```
 
    ```python
+   import json
+
    from nmdc_metadata_suggestor_ai_tool.llm_client import LLMClient
    from nmdc_metadata_suggestor_ai_tool.recommendation_pipeline import run_recommendation_pipeline
 
-   submission_object = {
-       # NMDC submission JSON payload
-   }
+   # Any NMDC submission JSON payload. This example fixture ships with the repo:
+   # a soil study with a data DOI, so the run also exercises DOI abstract ingestion.
+   with open("tests/fixtures/test_submission.json") as f:
+       submission_object = json.load(f)
 
    client = LLMClient(access_provider="gcp")
    result = run_recommendation_pipeline(submission_object, client)
-   print(result.model_dump())
+   print(result.model_dump_json(indent=2))
    ```
+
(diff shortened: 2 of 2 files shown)


.env.example
```diff
@@ -29,6 +29,13 @@ VERTEX_PROJECT_ID=your_gcp_project_id
 # Default Vertex region (the code default is us-east5 if unset)
 GCP_REGION=us-east5
 
+# Claude Agent SDK configuration, used only by ConversationManager.agentic().
+# That path runs a Claude model through the Claude Code CLI, which reads these to
+# route to Vertex. Unused by the Gemini path.
+CLAUDE_CODE_USE_VERTEX=1
+ANTHROPIC_VERTEX_PROJECT_ID=your_gcp_project_id
+CLOUD_ML_REGION=us-east5
```
hesspnnl: Is `CLOUD_ML_REGION` actually needed? I do not have that set in my local .env file and we recently got rid of all the `CLOUD_ML_REGION` artifacts in [this PR](https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/96#discussion_r3559939824)

README.md
```diff
@@ -24,6 +24,7 @@ Environment variables used by `LLMClient` and `ConversationManager`:
 - `GOOGLE_APPLICATION_CREDENTIALS`: Path to a GCP service account JSON file (for Vertex AI).
 - `VERTEX_PROJECT_ID`: (Optional) GCP project id for Vertex. If not provided, the SDK will attempt to infer it from credentials.
 - `GCP_REGION`: (Optional) Vertex region override for Gemini calls (falls back to `GCP_REGION`, then `us-east5`).
+- `CLAUDE_CODE_USE_VERTEX`, `ANTHROPIC_VERTEX_PROJECT_ID`, `CLOUD_ML_REGION`: Required only by the agentic path (`ConversationManager.agentic()`), which runs a Claude model through the Claude Code CLI. Without them the CLI does not route to Vertex, and the run fails reporting that the model may not exist or is not accessible.
```
hesspnnl:
```suggestion
- `CLAUDE_CODE_USE_VERTEX` (Required by ConversationManager.agentic()) : Forces `claude_agent_sdk` to use vertex AI for credentials. Required for our use as we use GCP for agentic auth. 
```
Something more explicit on what each one does would be more beneficial! doesnt need to be these exact words.

.env.example
```diff
@@ -29,6 +29,13 @@ VERTEX_PROJECT_ID=your_gcp_project_id
 # Default Vertex region (the code default is us-east5 if unset)
 GCP_REGION=us-east5
 
+# Claude Agent SDK configuration, used only by ConversationManager.agentic().
+# That path runs a Claude model through the Claude Code CLI, which reads these to
+# route to Vertex. Unused by the Gemini path.
+CLAUDE_CODE_USE_VERTEX=1
+ANTHROPIC_VERTEX_PROJECT_ID=your_gcp_project_id
```
hesspnnl: Is `ANTHROPIC_VERTEX_PROJECT_ID` also required? I do not have that one set in my local .env either.

hesspnnl (changes requested):
Thanks for adding these! I have a couple questions as our env vars are slightly different.

.env.example
sujaypatil96: Good catch! this was my agent leading me astray. `CLOUD_ML_REGION` isn't needed, i'll remove it from this PR.

sujaypatil96: Looks like `us-east5` is what the region resolution defaults in the `claude-agent-sdk` package.

.env.example
sujaypatil96: Good catch again! you're right - looks like `ANTHROPIC_VERTEX_PROJECT_ID` isn't required either. The service account JSON file that we're pointing the `GOOGLE_APPLICATION_CREDENTIALS` env var to has a `project_id` which is what the CLI will pull out from that file. So adding `ANTHROPIC_VERTEX_PROJECT_ID` technically wouldn't do us any harm, but why explicitly restate things when we don't have to. I'll go ahead and remove it.

hesspnnl (approved):
looks great. thanks for double checking those env vars

PR #153, opened 2026-08-24
Declare every target in .PHONY, and check it stays that way

### LINKED ISSUE #115: Complete .PHONY declarations in the Makefile
A Copilot review on #107 noted the `Makefile`'s `.PHONY` line omits several non-file targets: `all-install`, `prod-install`, `docker-dev-build`, `docker-dev-down`, `docker-shell`, and `run`. If a file with one of those names ever exists, `make` may treat the target as up to date and skip it.

Left out of #107 to keep that PR scoped to dependency tooling. Worth completing separately.

Note before doing this: several of the gaps are `docker-*` targets. #104 is changing the docker files and #85 proposes removing docker artifacts entirely, so coordinate with those first. It may be simplest to complete `.PHONY` after the docker direction settles, and drop any targets that get removed.

turbomam:
## Why

Closes https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/115.

The `.PHONY` line listed 17 names while the Makefile defines 23 targets. A phony target that is
not declared gets skipped whenever a file of the same name exists, and make says
`make: 'run' is up to date.` rather than failing, so the skip is silent.

The issue asked to wait for the docker direction to settle first. It has:
https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/104 merged on 2026-07-10.

## What changed

Six real targets added to `.PHONY`: `all-install`, `prod-install`, `docker-dev-build`,
`docker-dev-down`, `docker-shell`, `run`. One name removed, `lint-fix`, which is not a target.

The rule used: a target is phony when running it produces no file named after the target. All 23
targets qualify. None writes an artifact of its own name, and no target name collides with a path
in the repo.

Second commit adds `make check-phony`, called by `make lint`, so the line cannot drift quietly
again. It fails on a target missing from `.PHONY`, a `.PHONY` entry that is not a target, a target
name that exists as a path, or a duplicate target.

## Validation

Behaviour, not just text. With files named `run`, `all-install` and `docker-shell` planted in the
tree, `make -n <target>`:

| target | before | after |
|---|---|---|
| `run` | `make: 'run' is up to date.` | `uv run nmdc-suggestor` |
| `all-install` | `make: 'all-install' is up to date.` | `uv sync` |
| `docker-shell` | `make: 'docker-shell' is up to date.` | `docker-compose exec app bash` |

The check was tested against each failure mode it claims to catch (undeclared target, stale
`.PHONY` entry, target name existing as a path) and fails on each.

`make lint` clean: ruff, mypy on 51 source files, and the new check. `make test`: 257 passed,
57 deselected.

## On checkmake

checkmake is the obvious off-the-shelf linter for this and it ships a rule named `phonydeclared`,
so it looks like it would have caught this. It would not. From its source, the condition is
`len(rule.Body) == 0 && !ok`: it only fires on targets with no recipe body. Every target here has
one, so checkmake reports zero violations against the exact file this PR fixes. Its other phony
rule, `minphony`, only checks that a configured list (default `all,clean,test`) is present.

That is why this is a local check rather than a new dependency. Recorded in
`docs/makefile-phony.md` so the evaluation is not repeated.

## Documentation impact

New `docs/makefile-phony.md`: the rule, what the check enforces, and what to do when adding a
target.

## Scope and complexity

Small. One line of Makefile plus a target, an 82-line script with no dependencies, and a doc.

Three broken targets turned up while verifying this, all pre-existing and left alone here to keep
the PR reviewable. Filed separately:

- https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/154 the five DOI targets use the pre-rename package path `nmdc_metadata_suggestor`
- https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/155 six `docker-*` targets reference files that PR 104 deleted
- https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/156 `run` calls a console script that `pyproject.toml` does not define

Completing `.PHONY` is what makes 155 and 156 reachable, since those targets were being skipped as
up to date.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,15 @@
-# Targets that really do build a file or directory named after themselves.
-# These must NOT be listed in .PHONY; every other target must be. Empty today,
-# because nothing here builds an artifact of its own name. See
-# docs/makefile-phony.md.
+# Every target below is listed in .PHONY, because none of them creates a file
+# named after itself. That is the whole rule: declare a target phony unless
+# running it writes a file or directory of that exact name.
+#
+# FILE_TARGETS is the escape hatch for the exception. If a target ever does
+# build a file named after itself, list it here INSTEAD of in .PHONY, and make
+# will use its timestamp to skip rebuilds. Nothing qualifies today, so it is
+# empty. `make check-phony` fails if a target is in both lists or neither.
+# Background: docs/makefile-phony.md.
 FILE_TARGETS =
 
-.PHONY: help all-install prod-install dev-install test test-integration lint check-phony format security check-deps clean docker-build docker-dev-build docker-run docker-dev docker-dev-down docker-shell run validate-doi classify-doi classify-fixture get-abstract get-abstracts
+.PHONY: help all-install prod-install dev-install test test-integration lint check-phony format security check-deps clean run validate-doi classify-doi classify-fixture get-abstract get-abstracts
 
 help: ## Show this help message
 	@echo "Usage: make [target]"
@@ -59,24 +64,6 @@ clean: ## Clean up generated files
 	rm -rf *.egg-info
 	find . -type d -name __pycache__ -exec rm -rf {} +
 	find . -type f -name "*.pyc" -delete
-
-docker-build: ## Build production Docker image
-	docker build -t nmdc-suggestor:latest .
-
-docker-dev-build: ## Build development Docker image
-	docker build -f Dockerfile.dev -t nmdc-suggestor:dev .
-
-docker-run: ## Run production Docker container
-	docker run --env-file .env nmdc-suggestor:latest
-
-docker-dev: ## Start development environment with Docker Compose
-	docker-compose up
-
-docker-dev-down: ## Stop development environment
-	docker-compose down
-
-docker-shell: ## Open shell in development container
-	docker-compose exec app bash
 
 run: ## Run the application locally
 	uv run nmdc-suggestor
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help all-install prod-install dev-install test test-integration lint format security check-deps clean docker-build docker-dev-build docker-run docker-dev docker-dev-down docker-shell run validate-doi classify-doi classify-fixture get-abstract get-abstracts
+.PHONY: help all-install prod-install dev-install test test-integration lint check-phony format security check-deps clean docker-build docker-dev-build docker-run docker-dev docker-dev-down docker-shell run validate-doi classify-doi classify-fixture get-abstract get-abstracts
 
 help: ## Show this help message
 	@echo "Usage: make [target]"
@@ -23,7 +23,11 @@ test-integration: ## Run integration tests against real APIs
 
 lint: ## Run linters
 	uv run ruff check
-	uv run mypy src 
+	uv run mypy src
+	$(MAKE) check-phony
+
+check-phony: ## Verify .PHONY lists exactly the targets the Makefile defines
+	uv run python scripts/check_phony.py
 
 format: ## This is a two part command that both checks, then auto-fixes formatting issues where it can.
 	uv run ruff format
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: help dev-install test test-integration lint lint-fix format security check-deps clean docker-build docker-run docker-dev validate-doi classify-doi classify-fixture get-abstract get-abstracts
+.PHONY: help all-install prod-install dev-install test test-integration lint format security check-deps clean docker-build docker-dev-build docker-run docker-dev docker-dev-down docker-shell run validate-doi classify-doi classify-fixture get-abstract get-abstracts
 
 help: ## Show this help message
 	@echo "Usage: make [target]"
--- a/Makefile
+++ b/Makefile
@@ -1,3 +1,9 @@
+# Targets that really do build a file or directory named after themselves.
+# These must NOT be listed in .PHONY; every other target must be. Empty today,
+# because nothing here builds an artifact of its own name. See
+# docs/makefile-phony.md.
+FILE_TARGETS =
+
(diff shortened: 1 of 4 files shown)


scripts/check_phony.py
```diff
@@ -0,0 +1,82 @@
...
+    # The other half of the rule: a target that really does build a file of
+    # its own name must not be phony. We cannot read recipes, but a target
+    # name existing as a path is the signal worth failing on.
+    root = MAKEFILE.parent
+    collide = [t for t in targets if (root / t).exists()]
+    if collide:
+        problems.append(
+            "target names that exist as a path, so they may build a file of "
+            f"their own name and should not be phony: {' '.join(collide)}"
+        )
```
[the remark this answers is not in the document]
turbomam: Right, and reworded in ecc2842. The message now says the name also exists as a path in the working tree, tracked or not, and no longer implies the target built it.

Keeping the check itself. An untracked stray directory named `run` skips the target exactly as a committed one would, so failing on it is the behaviour I want. The message was the part that was wrong.

docs/makefile-phony.md
```diff
@@ -0,0 +1,55 @@
...
+Every target in this Makefile is phony today. None of them writes a file named after itself.
+
+## Why this is checked automatically
+
+`make lint` runs `make check-phony`, which runs `scripts/check_phony.py`. It fails on:
+
+- a target missing from `.PHONY`
+- a `.PHONY` entry that is not a target
+- a target name that exists as a path in the repo, which suggests it should not be phony
+- a target defined twice
```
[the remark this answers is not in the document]
turbomam: Fixed in ecc2842. The doc now says "in the working tree, tracked or untracked", and adds a line telling you to delete the stray path if `make lint` trips on it locally, which is the confusing case you are pointing at.

docs/makefile-phony.md
```diff
@@ -0,0 +1,60 @@
...
+present, and says nothing about the rest of the file.
+
+So a green `checkmake` run is not evidence that `.PHONY` is complete. That is why this repo uses
+a local check instead of adding a dependency.
+
+## Adding a target
+
+Add it to `.PHONY` in the same place. `make check-phony` fails if you forget. If the new target
+genuinely builds a file named after itself, rename the target after the path it builds and leave
+it out of `.PHONY`.
```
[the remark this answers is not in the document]
turbomam: Good catch, and it was a real contradiction rather than a doc problem. The check demanded every target appear in `.PHONY`, so a target that legitimately builds a file of its own name got reported twice: once for being absent from `.PHONY`, where by the rule it does not belong, and again for colliding with the path it is supposed to create. The doc told you to do the thing the check rejected.

Fixed in d0c9a80 by making the second half of the rule declarable. Targets like that go in a `FILE_TARGETS` variable at the top of the Makefile, which keeps them out of `.PHONY` and exempts them from the path check, since the path existing is the expected outcome there. Listing a target in both is now its own error, as is naming something in either that is not a target.

`FILE_TARGETS` is empty today because nothing here builds an artifact of its own name. I tested the case you described end to end: the target fails the check when it is in neither list, passes once declared in `FILE_TARGETS` even with the file present, and fails if it is in both.

Makefile
```diff
@@ -1,4 +1,10 @@
-.PHONY: help dev-install test test-integration lint lint-fix format security check-deps clean docker-build docker-run docker-dev validate-doi classify-doi classify-fixture get-abstract get-abstracts
+# Targets that really do build a file or directory named after themselves.
+# These must NOT be listed in .PHONY; every other target must be. Empty today,
+# because nothing here builds an artifact of its own name. See
+# docs/makefile-phony.md.
+FILE_TARGETS =
+
+.PHONY: help all-install prod-install dev-install test test-integration lint check-phony format security check-deps clean docker-build docker-dev-build docker-run docker-dev docker-dev-down docker-shell run validate-doi classify-doi classify-fixture get-abstract get-abstracts
```
hesspnnl: lets remove docker artifacts

Makefile
```diff
@@ -1,4 +1,10 @@
-.PHONY: help dev-install test test-integration lint lint-fix format security check-deps clean docker-build docker-run docker-dev validate-doi classify-doi classify-fixture get-abstract get-abstracts
+# Targets that really do build a file or directory named after themselves.
```
hesspnnl: these comments are confusing to me - are they saying that we need to declare targets in phony or we should not declare them?

hesspnnl (changes requested):

Makefile
turbomam: Done in 6e0fb73, and thank you for deciding it. All six are gone.

They were worse than redundant: every one of them failed, because #104 deleted the Dockerfiles and the compose file back in July and left the targets behind. Four of the six were also missing from `.PHONY`, so they were being reported as "up to date" and skipped rather than failing visibly.

Also removed the `docker-compose up` block from `CONTRIBUTING.md`, which told a new contributor to run a compose file that has not existed for six weeks. That is the same artifact, so it seemed to belong with your decision rather than in a separate issue. This closes https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/155.

Makefile
turbomam: Fair, the comment was backwards. It opened by describing the exception, so it read as a claim about every target. Rewritten in 6e0fb73.

To answer directly: **declare a target in `.PHONY` unless running it writes a file or directory of that exact name.** Every target in this Makefile qualifies, so all 18 are listed.

`FILE_TARGETS` is only the escape hatch for the exception. If a target ever does build a file named after itself, it goes there *instead of* `.PHONY`, so make can use the file's timestamp to skip a rebuild. Nothing qualifies today, so it is empty. `make check-phony` fails if a target ends up in both lists or in neither.

The new comment says that in that order, rule first.

hesspnnl (approved):
thanks for the changes. that makes much more sense

PR #157, opened 2026-08-24
Type-check tests as well as src

### LINKED ISSUE #102: Type-check tests, not just src (extend mypy scope)
CI runs `uv run mypy src`, so test files are not type-checked. That scope came in when the Copilot coding agent scaffolded `ci.yml` (2026-02-03), not as a deliberate exemption. The tests are already written in a typed style and `disallow_untyped_defs = true` is set, so extending to `uv run mypy src tests` would likely pass with little change and catch type errors in test code. Low effort, optional.

turbomam:
## Why

Closes https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/issues/102, scoped by
@hesspnnl on 2026-08-21 as "add the test dir to type checking, update tests as needed".

## What changed

mypy's scope now includes `tests` in three places: the `Makefile` lint target, `CONTRIBUTING.md`,
and `.github/workflows/ci.yml`.

The CI one is the change that actually does anything. `ci.yml` calls `uv run mypy src` directly
rather than going through `make lint`, so editing only the Makefile would have left CI checking
`src` alone and the scope extension would have been cosmetic.

Turning it on produced 48 errors, all in `tests`. `src` was already clean. They are fixed rather
than suppressed: nothing was added to an ignore list, no rule was relaxed, and
`disallow_untyped_defs` still applies to tests the same as to source.

## The fixes, by shape

- **Fixture loaders** declared `-> dict` and returned `json.load(f)`, which is `Any`. They now
  declare `-> dict[str, Any]` and annotate the loaded value.
- **`.get()` chains on fixture keys that are always present** became direct indexing. `.get()`
  returns `Any | None` and every one of these was immediately dereferenced, so the `Optional` was
  never real.
- **Optionals that tests dereference** are asserted non-None first. This also makes the assumption
  visible in the failure output when a fixture changes shape.
- **`_FakeLLMClient`** is cast to `LLMClient` at the seven call sites rather than subclassing it,
  so the fake does not inherit `LLMClient`'s credential setup. A comment at the class says so.
- **Missing annotations** on pytest fixture parameters (`tmp_path`, `monkeypatch`) and two inner
  helpers.

## Validation

`ruff check`, `ruff format --check` on 87 files, `mypy src tests` on 69 source files, and
`pytest`: 257 passed, 57 deselected. Same 257 as before the change.

## Documentation impact

`CONTRIBUTING.md`'s type-check command updated to match.

## Scope and complexity

Mechanical. 87 insertions, 53 deletions across 13 test files plus three one-line config changes.
No source file touched.

Note for whoever merges second: this and
https://github.com/microbiomedata/nmdc-metadata-suggestor-ai-tool/pull/153 both edit the
`Makefile` lint target, on adjacent lines. Either may need a trivial rebase.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,7 @@ jobs:
         uv run ruff format --check
 
     - name: Run type checking
-      run: uv run mypy src
+      run: uv run mypy src tests
 
     - name: Check dependency hygiene
       run: make check-deps
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -55,7 +55,7 @@ make test
 # Or individually:
 uv run ruff format     # Format code
 uv run ruff check      # Lint code
-uv run mypy src        # Type check
+uv run mypy src tests  # Type check
 uv run pytest          # Run tests
 ```
 
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ test-integration: ## Run integration tests against real APIs
 
 lint: ## Run linters
 	uv run ruff check
-	uv run mypy src
+	uv run mypy src tests
 	$(MAKE) check-phony
 
 check-phony: ## Verify .PHONY lists exactly the targets the Makefile defines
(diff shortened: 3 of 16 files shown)


hesspnnl (approved):
seems good. thanks!

PR #163, opened 2026-08-28
129 add award doi api

JamesTessmer:
Resolves #129 

This PR adds functionality to the `osti.py` file by implementing OSTI award API querying.
This implementation mimics the existing API querying code.

To make this functionality usable, supporting changes in `main.py` and `constants.py` were added
- Adding a fetcher 
- Adding `osti_award` to `_SOURCE_FETCHERS`
- Adding the OSTI award API url as a constant

This PR is in progress, things left to do:
- Ensure the added changes are enough for functionality
- Once that's done, update test files

### DIFF
--- a/src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
@@ -27,8 +27,7 @@ from nmdc_metadata_suggestor_ai_tool.doi_ingestion.jgi import try_jgi
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.kbase import try_kbase
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.massive import try_massive
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.openalex import try_openalex
-from nmdc_metadata_suggestor_ai_tool.doi_ingestion.osti import try_osti
-from nmdc_metadata_suggestor_ai_tool.doi_ingestion.osti import try_osti_award
+from nmdc_metadata_suggestor_ai_tool.doi_ingestion.osti import try_osti, try_osti_award
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.pubmed import try_pubmed
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.zenodo import try_zenodo
 from nmdc_metadata_suggestor_ai_tool.models.doi import DoiClassification, SourceRetrievalResult
@@ -449,11 +448,14 @@ def _fetch_osti(
     """Fetch and wrap context from OSTI API."""
     return _fetch_resolver_context(doi, provider, "osti", attempts, source_errors, try_osti)
 
+
 def _fetch_osti_award(
     doi: str, provider: str | None, attempts: list[str], source_errors: SourceErrors
 ) -> SourceRetrievalResult | None:
     """Fetch and wrap context from OSTI Award API."""
-    return _fetch_resolver_context(doi, provider, "osti_award", attempts, source_errors, try_osti_award)
+    return _fetch_resolver_context(
+        doi, provider, "osti_award", attempts, source_errors, try_osti_award
+    )
 
 
 _SOURCE_FETCHERS: dict[str, Fetcher] = {
--- a/src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
@@ -486,5 +486,4 @@ _PROVIDER_API_SOURCES = {
     "cyverse",
     "zenodo",
     "osti",
-    "osti_award",
 }
--- a/src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
+++ b/src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
@@ -28,6 +28,7 @@ from nmdc_metadata_suggestor_ai_tool.doi_ingestion.kbase import try_kbase
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.massive import try_massive
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.openalex import try_openalex
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.osti import try_osti
+from nmdc_metadata_suggestor_ai_tool.doi_ingestion.osti import try_osti_award
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.pubmed import try_pubmed
 from nmdc_metadata_suggestor_ai_tool.doi_ingestion.zenodo import try_zenodo
 from nmdc_metadata_suggestor_ai_tool.models.doi import DoiClassification, SourceRetrievalResult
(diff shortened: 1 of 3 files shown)


src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
JamesTessmer, line 110: I feel like this block might need an update, but I'm not sure exactly how we want to do that. Right now the supporting functions are kind of set up with a binary 'is it a publication DOI or is it a data DOI'. Do we fold the award DOI into either of those? Or do we want to do a little bit of a bigger lift to check for awards?

As it is now, I'm not sure that the `_fetch_osti_award` func is called if an award DOI is given without `osti_award` explicitly being given in the `sources` variable.

caufieldjh (approved):
~~I think this should work as a start - I will add some comments but don't consider any of them blockers.~~ Oops, I take that back. I testes and there are some blockers.

src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/osti.py
```diff
@@ -112,3 +115,59 @@ def query_osti_by_doi(osti_doi: str) -> list[dict]:
...
+        return None
+
+    if not data or len(data) == 0:
+        append_error(errors, f"No award found for award DOI: {award_doi}")
+        return None
+
+    description = None
+
+    # When querying an award DOI there should only be one record
+    raw_description = data[0].get("description")
```
caufieldjh: This line encounters a `KeyError` on the live API, since it doesn't return anything list-like. The E-Link 2 API *does* return a list, but the award API returns a Solr envelope that looks more like:
```
{
  "responseHeader": {"status": 0, "params": {"fq": "award_doi:(10.46936/...)"}},
  "response": {"numFound": 1, "start": 0, "docs": [{ ...award... }]}
}
```
`ValueError` and `RequestException` are caught just fine, but this isn't one of those, so it breaks.

caufieldjh: Also: `description` may be a valid field but live records look like they use `award_description` (and `award_title`) instead.
I think retrieving these fields from `data["response"]["docs"]` should work.

src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
```diff
@@ -465,6 +474,7 @@ def _fetch_osti(
     "openalex": _fetch_openalex,
     "pubmed": _fetch_pubmed,
     "osti": _fetch_osti,
+    "osti_award": _fetch_osti_award,
 }
 
 _PROVIDER_API_SOURCES = {
```
caufieldjh: Should osti_award go here, too?

caufieldjh (changes requested):
Tests would be very nice to have!

src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/osti.py
JamesTessmer: Good catch, I think they must have updated that very recently. Their documentation page still supports just a `description` field but that's not what's actually being returned.

I had to talk to someone at OSTI to clear up some things since their documentation was not correct on the permissions needed to use the API as well, so it seems like they may just need to update the docs in general.

src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
JamesTessmer: This was one thing I was unsure of. I wasn't sure if I should add it to `_PROVIDER_API_SOURCES` or what changes would be necessary to make the driving function recognize the new award api functions.

It seems like a lot of the code is built around DOIs either being publication or data DOIs.
I wasn't sure if we wanted to lump this into publications (which seems like the more similar of the two) or if we wanted to do something separate.

Mostly wondering if there are cases where `default_source_order` would not catch the award API.

JamesTessmer:
@caufieldjh I've pushed up the changes requested, should be good for another review

src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
caufieldjh: Yes, the existing code does assume that a DOI will either refer to a publication or a data object, but we can capture award DOIs too with a few changes.
One spot is in `_check_classification_gate()` - it doesn't accept `award_doi` as a valid option.
The other is in `default_source_order`, like you're wondering about. You could check for `if is_award` before the `if is_publication` check.

src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
```diff
@@ -478,4 +488,5 @@ def _fetch_osti(
     "cyverse",
     "zenodo",
     "osti",
+    "osti_award",
```
caufieldjh: Great, this should definitely be here, but right now it doesn't do anything because `TARGET_PROVIDER_PREFIXES` in `constants.py` doesn't map any DOI prefix to `osti_award`. I'm not 100 percent sure what prefix to use here; `10.46936` is used elsewhere as an example but I think that's specific to JGI award DOIs.

caufieldjh:
You'll want to verify this, but it looks like DOIs not returning an `award_description` get rejected by the resolver, so it returns `None` even if there are other valid fields, like an `award_title`.

src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
JamesTessmer: Yeah that was one thing I had noticed as well. It seems to me that the prefix depends on the provider and is not linked to OSTI. All the ones I had seen by looking at data in the submission portal pointed to places like JGI or EMSL and overlapped with existing values in that enum.

JamesTessmer:
> You'll want to verify this, but it looks like DOIs not returning an `award_description` get rejected by the resolver, so it returns `None` even if there are other valid fields, like an `award_title`.

Ah good catch, I hadn't considered that case. I guess this is a question for you and @hesspnnl is there more that we want to capture than the description? And if there is no description but there is other data do we want to return a more specific error or value that says something like, 'award present, but no description provided'?

src/nmdc_metadata_suggestor_ai_tool/doi_ingestion/main.py
JamesTessmer: Poking around more on their FAQ page I saw this https://www.osti.gov/award-doi-service/faqs#sec_what-metadata-fields-are-available-to-describe-award-dois

It looks like the prefix is site dependent, but there should be an infix for award DOIs.

hesspnnl:
> > You'll want to verify this, but it looks like DOIs not returning an `award_description` get rejected by the resolver, so it returns `None` even if there are other valid fields, like an `award_title`.
> 
> Ah good catch, I hadn't considered that case. I guess this is a question for you and @hesspnnl is there more that we want to capture than the description? And if there is no description but there is other data do we want to return a more specific error or value that says something like, 'award present, but no description provided'?

looking at the response shape [here](https://www.osti.gov/api/docs/adsapi/records/get-record) award title and description seem like the only fields of interest. i agree with harry we would want to return the value if either of those exist

